Issue #8, opened 2021-01-13
Create Sphinx documentation and host on Read The Docs


felix-hilden:
From #7: Let's build and host the documentation on RTD then! I'd be happy to help in this way, since for me it seems like low-hanging fruit 👍

So what would you like in the documentation, since *beartype* has no API?

I would propose the following, which I did for my library ([RTD site](https://tekore.readthedocs.io/en/stable/) for reference):
- RTD becomes the home site for *beartype*, because there is no ugly file contents blocking the readme, like on GitHub
- We move the changelog and the (extremely) terse reference documentation there into one section, and create another section for tutorials, examples, explanations and such. GitHub readme becomes essentially an introduction to the *repository*, not the package. Although at this stage there's an argument to be made for keeping things in the GitHub readme still.

Thoughts?

leycec:
**Yup.** This is the way. <sup>*Thanks, Mandalorian.*</sup>

I at least had the marginal foresight to format our current `README` as reStructuredText (reST). So, this should be a less painful transition than the [blood-dimmed tide of anarchy](https://poets.org/poem/second-coming) I'd initially envisioned. I just need to start toying around with [Sphinx](https://www.sphinx-doc.org/en/master) on my local geriatric devbox mostly.

On a natural scale of **[1—11]** where **1** is *"Do everything else first..."* and **11** is *"Do this now or the Baby Yoda gets it!"*, where would you say this personally lands in importance for you?

It goes without saying that all future @beartype projects like the proposed `deeptyping` and `pepitup` will use an RTD-hosted setup from the get-go.

felix-hilden:
Neat! Well, I could certainly lend a few hours for it tomorrow, which should be enough to at least some version up, so let's say a 7/11 😉 I can't remember if RTD requires write access to the repo. It shouldn't though, so once the PR is merged, the documentation will begin building. The appearance can be checked with local builds anyway, it's more so a matter of configuring RTD. Until tomorrow, then!

leycec:
**Wow.** Our first prospective PR! The excitement just keeps building.

I will, of course, promptly merge pretty much *anything* you propose. Since we currently have nothing, anything is already a distinct improvement worthy of everyone's praise and gratitude. :partying_face:

leycec:
[...]

PR #9, opened 2021-01-14
Documentation configuration

felix-hilden:
From #8: Here's a suggestion for RTD+Sphinx configuration.

RTD can be set up without any further configuration files from their UI. Would you still like it to use the YAML? I think it's simply a matter of preference.

When creating a RTD site, one can import from their GitHub account, or manually set up a site. It could be better if you create the site, and then if I'm needed, I can be added to the project. I'm not sure if manual creation has some drawbacks, but I'd prefer not to test it just to see that it does 😅

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,0 +1,2 @@
+sphinx==3.4.3
+sphinx-rtd-theme==0.5.1
--- a/doc/source/404.rst
+++ b/doc/source/404.rst
@@ -1,0 +1,8 @@
+========
+Beartype
+========
+Oops! The page you are looking for was not found.
+Maybe you'll find what you're looking for by searching the documentation
+or returning to the `home page <rtd_>`_.
+
+.. _rtd: https://beartype.rtfd.org
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -76,6 +76,9 @@ extensions = [
     # each of this project's Python modules and linking external references to
     # those modules to these listings.
     'sphinx.ext.viewcode',
+
+    # 3rd party Read The Docs HTML theme for neat and mobile-friendly doc site
+    'sphinx_rtd_theme',
 ]
 
 # ....................{ EXTENSIONS ~ napoleon             }....................
@@ -95,7 +98,7 @@ napoleon_google_docstring = False
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = 'alabaster'
+html_theme = 'sphinx_rtd_theme'
 
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,


doc/requirements.txt
```diff
@@ -0,0 +1,2 @@
+sphinx==3.4.3
+sphinx-rtd-theme==0.5.1
```
felix-hilden, line 2: These are just for RTD. How would you like to declare documentation dependencies? In `beartype.meta` as a separate category like `LIBS_DEVELOPMENT`?

leycec:
**Wowza!** You even hooked us up with:

* A saner Sphinx theme replicating the rendering of our documentation under RTD.
* A fun 404 error page redirecting back to our top-level domain (TLD), which [hilariously](http://acronymsandslang.com/acronym_image/2615/e4e09fa45f010f68e43698cfcc7b9ce6.jpg) appears to be `beartype.rtfd.org`. I can't believe `rtfd.org` is an actual alias for the canonical domain `readthedocs.org`... *but it is.*

doc/requirements.txt
leycec: **Right.** This is the standard approach and for good reason. But I'm a dependency wierdo, as you've probably already observed from `beartype.meta`.

The `doc/requirements.txt` approach is *absolutely* fine. We can run with this as is. My personal preference, though, would be to define a top-level `.readthedocs.yml` configuration file resembling:

```
version: 2

python:
  version: 3.7
  install:
    - method: pip
      path: .
      extra_requirements:
        - doc
  system_packages: true
```

RTD would then install `beartype` as:

```
$ pip install .[docs]
```

This is kinda nice, because it enables end users to do the same. Maybe someone wants that? Who knows! But I'd like to give that to them, if we can.

To facilitate this, I'll need to do a bit of additional legwork on my end. But that's fine! I have legs, so legwork is feasible. This includes:

* Defining a new `beartype.meta.LIBS_DOCTIME_MANDATORY` tuple resembling:

```
LIBS_DOCTIME_MANDATORY = {
    'sphinx >=3.4.3',
    'sphinx-rtd-theme >=0.5.1',
}
```

* Defining a new `'extras_require'` key-value pair in `setup` resembling:

```
        'doc': meta.LIBS_DOCTIME_MANDATORY,
```

Don't feel to obliged to do *any* of that. That's entirely my obligation.

Here's where you come in! My main question is the pinning of dependencies in `doc/requirements.txt`. I usually like to avoid pinning in open-source projects to:

* Improve forward compatibility with a wider range of versions.
* Enable end users to install dependencies with their preferred package manager (e.g., `apt`), which often fail to satisfy specific versions.
* Simplify my life by notifying me with [failing badges](https://docs.readthedocs.io/en/stable/badges.html) and automated e-mails when dependency bumps break our documentation build.

Does all of that make sense or am I inviting a horrifying reenactment of Pompeii by failing to pin versions?

felix-hilden: Yep! I mostly agree. In my mind pinning exact versions for RTD is more like a pin for deployment, rather than the library, for which I wholeheartedly second more loose requirements. But perhaps `sphinx ~= 3.4` and such could be appropriate instead of simply `>=`? Your call! And having users do `pip install -e .[docs,test]` vs `-e .[dev]` could be an argument to be had as well. [I've used](https://github.com/felix-hilden/tekore/blob/master/setup.py#L36) a `dev` collection to gather all the different extras into one, which seems to be common practice. But two isn't that bad. Three gets a bit annoying.

felix-hilden:
That *is* hilarius to me as well 😄 And of course, if we need it (or if you indeed *prefer* YAML, or just having the settings in the repo rather than UI), then the config file should be created. Currently on my project I've gotten away with simply specifying advanced settings, which can be used to define both the build requirements and conf.py location. Again, your call!

So if you'd like to not pin the exact versions, I think it would be better to use the config file, since we can simply use the library extra deps.

doc/requirements.txt
leycec: > In my mind pinning exact versions for RTD is more like a pin for deployment...

This makes rationality. On the one sweaty hand, duplicating requirements across `doc/requirements.txt` and `beartype/meta.py` [violates DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself); on the other sweaty hand, *not* pinning RTD requirements invites build failures for no particularly good reason.

*Ugh, choices.* O.K., I'm sold. Let's duplicate requirements as you say, because breaking documentation builds for no particularly good reason is word. Also, your word is law.

> And having users do `pip install -e .[docs,test]` vs `-e .[dev]` could be an argument to be had as well.

You're... right, actually. If I was a contributor, I'd personally be annoyed to the point of rabidly gnashing teeth at even two different extras. Sadly, the `test` extra is required by our top-level `tox` configuration:

```
extras =
    # Install all optional runtime dependencies to maximize test coverage.
    all

    # Install all mandatory test-specific dependencies. This is the official
    # solution supported by "tox" developers for eliminating redundancy between
    # testing dependencies listed within this file and the top-level "setup.py"
    # script. While non-intuitive, we have little recourse. See also:
    #     https://stackoverflow.com/questions/29870629/pip-install-test-dependencies-for-tox-from-setup-py
    #     https://stackoverflow.com/questions/39922650/tox-tests-use-setup-py-extra-require-as-tox-deps-source
    #     https://github.com/tox-dev/tox/issues/13#issuecomment-247788280
    # Note that this also requires ".[test]" to be listed as a dependency.
    test
```

*Ugh.* But that's fine. We can additionally define a new `dev` extra defined as the tuple union of `test` and `doc`, enabling contributors to trivially `pip install -e .[dev]` without the fist-clenching hassle of extra extras. For example, in `setup.py`:

        'dev': meta.LIBS_TESTTIME_MANDATORY + meta.LIBS_DOCTIME_MANDATORY,

Behold, the horrifying black magic of `beartype.meta`! :mage_man:

leycec:
This is great as is, actually. It's late, the cat is gently wheezing on the bed, and [my head feels like it's been trepanned](https://en.wikipedia.org/wiki/Trepanning). So, let's just merge this as is!

Thanks again for all the hot PR action, Felix. This Moai statue's for you. :moyai:

felix-hilden:
You're most welcome, and much appreciated!

Issue #18, opened 2021-01-30
Implementing `get_callable_filename_or_placeholder`


Heliotrop3:
## Task List
- [ ] Implement a new exception: `_BeartypeUtilCallableException`
- [ ] Implement `get_callable_filename_or_placeholder`
    - [ ] Handle built-in functions/methods
    - [ ] Raise an error when:
        - [ ] A non-callable is passed
   - [ ] Returns a string when: 
        - [ ] A function/method missing `.__code__.co_filename`
        - [ ] A class whose `__module__.__file__` is `None`
   - [ ] Return the absolute file path of the passed Callable
- [ ] Document  
- [ ] Unit-test

## Introductions
Taken from beartype.beartype._util.utilcallable

```
#FIXME: Implement us up
#FIXME: Unit test us up.
def get_callable_filename_or_placeholder(func: Callable) -> str:
    '''
    Absolute filename of the uncompiled Python script or module physically
    declaring the passed callable if any *or* the placeholder string
    ``"<string>"`` implying this callable to have been dynamically declared
    in-memory otherwise.

    Parameters
    ----------
    func : Callable
        Callable to be inspected.

    Returns
    ----------
    str
        Either:

        * If this callable is physically declared by an uncompiled Python
          script or module, the absolute filename of this script or module.
        * Else, the placeholder string ``"<string>"`` implying this callable to
          have been dynamically declared in-memory.

    See Also
    ----------
    :func:`inspect.getsourcefile`
        Inefficient stdlib function strongly inspiring this implementation,
        which has been highly optimized for use by the performance-sensitive
        :func:`beartype.beartype` decorator.
    '''
```
Note: `inspect.getfile` is heavily referenced throughout. It does the heavy lifting of `inspect.getsourcefile`

## Down the Rabbit Hole
As noted by in the official [documentation](https://docs.python.org/3/library/functions.html?#callable), `callable` can throw false positives. Curiosity and Google led to a StackOverflow post with a [comment](https://stackoverflow.com/questions/111234/what-is-a-callable#comment3252199_111255) declaring the best way check is EAFP. As a following comment points out, there was a brief interlude from Python 3.0 to 3.2 where `callable` was not in Python. It's reintroduction was, [according to the devs](https://bugs.python.org/issue7006#msg95400), a quality of life revert aimed at addressing long-distance calls: calls to objects far from their point of summoning. 

[...]

PR #19, opened 2021-02-06
feat: Add base for `utilcallable`

Heliotrop3:
Implements the basic grab of filename or placeholder for any callable
object. Reference the issue at the bottom for additional details.
Types were combined in cave.py both for  current and future use.

Breakdown of Changes:
 - Add a new exception `_BeartypeUtilCallableException`
 - Add a tuple to cave.py off all callable CType named `CallableCtypes`
 - Implement `get_callable_filename_or_placeholder`
 - Add test for extent of current implementation

Reference: #18

### DIFF
--- a/beartype/_util/utilcallable.py
+++ b/beartype/_util/utilcallable.py
@@ -19,11 +19,10 @@ This private submodule is *not* intended for importation by downstream callers.
 #submodules lying around. Huzzah!
 
 # ....................{ IMPORTS                            }....................
-import sys
-import os.path
+from sys import modules
 
 from collections.abc import Callable
-from beartype.cave import ClassType, CallableCTypes
+from beartype.cave import ClassType, CallableCTypes, MethodBoundInstanceOrClassType
 
 from beartype.roar import _BeartypeUtilCallableException
 # ....................{ GETTERS                           }....................
@@ -58,29 +57,28 @@ def get_callable_filename_or_placeholder(func: Callable) -> str:
     '''
     if callable(func) is False:
         raise _BeartypeUtilCallableException(
-            f"Attempted to pass non-callable:{repr(func)}")
+            f"Attempted to pass non-callable: {repr(func)}")
+    
+    filename = '<string>'
 
-   # Grab location of the module declaring class
-    elif isinstance(func, type):
+    # Grab location of the module declaring class.
+    if isinstance(func, type):
 
+        func_module_name = func.__module__
+        
         # Check the Class has a module thats registered with the system.
-        if (getattr(func,"__module__", None) and
-            sys.modules.get(func.__module__) and
-            getattr(sys.modules.get(func.__module__),"__file__", None)):
-            filename = os.path.abspath(sys.modules.get(func.__module__).__file__)
-        # Class was dynamically declared.
-        else:
-            filename = "<string>"
+        if func_module_name:
+            filename = getattr(modules[func_module_name],'__file__', filename)
+
     else:
         # Built-in types are callables with no .py source file.
-        if isinstance(func,CallableCTypes):
+        if isinstance(func, CallableCTypes):
             filename = "<builtin>"
-        # Check the function/method contains a code object with the declaring filename.
-        elif (getattr(func, "__code__", None) and
-              getattr(func.__code__, "co_filename", None)):
-            filename = os.path.abspath(func.__code__.co_filename)
-        # Callable was dynamically declared.
+
         else:
-            filename = "<string>"
+            if isinstance(func, MethodBoundInstanceOrClassType):
+                func = func.__func__
+
+            filename = func.__code__.co_filename
 
     return filename
--- a/beartype/_util/utilcallable.py
+++ b/beartype/_util/utilcallable.py
@@ -19,39 +19,68 @@ This private submodule is *not* intended for importation by downstream callers.
 #submodules lying around. Huzzah!
 
 # ....................{ IMPORTS                            }....................
-# from collections.abc import Callable
+import sys
+import os.path
 
+from collections.abc import Callable
+from beartype.cave import ClassType, CallableCTypes
+
+from beartype.roar import _BeartypeUtilCallableException
 # ....................{ GETTERS                           }....................
-#FIXME: Implement us up.
-#FIXME: Unit test us up.
-# def get_callable_filename_or_placeholder(func: Callable) -> str:
-#     '''
-#     Absolute filename of the uncompiled Python script or module physically
-#     declaring the passed callable if any *or* the placeholder string
-#     ``"<string>"`` implying this callable to have been dynamically declared
-#     in-memory otherwise.
-#
-#     Parameters
-#     ----------
-#     func : Callable
-#         Callable to be inspected.
-#
-#     Returns
-#     ----------
-#     str
-#         Either:
-#
-#         * If this callable is physically declared by an uncompiled Python
-#           script or module, the absolute filename of this script or module.
-#         * Else, the placeholder string ``"<string>"`` implying this callable to
-#           have been dynamically declared in-memory.
-#
-#     See Also
-#     ----------
-#     :func:`inspect.getsourcefile`
-#         Inefficient stdlib function strongly inspiring this implementation,
-#         which has been highly optimized for use by the performance-sensitive
-#         :func:`beartype.beartype` decorator.
-#     '''
-#
-#     pass
+def get_callable_filename_or_placeholder(func: Callable) -> str:
+    '''
+    Absolute filename of the uncompiled Python script or module physically
+    declaring the passed callable if any *or* the placeholder string
+    ``"<string>"`` implying this callable to have been dynamically declared
+    in-memory otherwise.
+
+    Parameters
+    ----------
+    func : Callable
+        Callable to be inspected.
+
+    Returns
+    ----------
+    str
+        Either:
+
+        * If this callable is physically declared by an uncompiled Python
+          script or module, the absolute filename of this script or module.
+        * Else, the placeholder string ``"<string>"`` implying this callable to
+          have been dynamically declared in-memory.
+
+    See Also
+    ----------
+    :func:`inspect.getsourcefile`
+        Inefficient stdlib function strongly inspiring this implementation,
+        which has been highly optimized for use by the performance-sensitive
+        :func:`beartype.beartype` decorator.
+    '''
+    if callable(func) is False:
+        raise _BeartypeUtilCallableException(
+            f"Attempted to pass non-callable:{repr(func)}")
+
+   # Grab location of the module declaring class
+    elif isinstance(func, type):
+
+        # Check the Class has a module thats registered with the system.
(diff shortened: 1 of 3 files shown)


Heliotrop3:
After spending ***WAY*** to long trying to figure out git merges and conflict resolution, I finally have it: a bald head. Once i git gud with workflows and ad the like, this won't be so agonizing.

leycec:
> After spending *WAY* to long trying to figure out git merges and conflict resolution, I finally have it: a bald head.

**Rejoice!** Due to `git`-ting gud, you too now know the joy of a reflective pate. We are bald brothers in code monkhood. :bald_man: 

I humbly apologize for generating suffering in your world. You are surely delighted to know that conflict resolution gets easier with time. Sadly, it never becomes joyful – merely tolerable. I still cringe whenever I see repetitions of greater-than or less-than signs.

`>>>>>>>>>>>>>>>`

**Triggered.**

You are also surely delighted to know that interactive rebasing (i.e., `git rebase -i`) is the most powerful artifact in the ancient `git` workflow. Once you know interactive rebasing, you know `git`. This is the maximum level of expertise a CLI guru can achieve. It's a high plateau, but it's all there is.

beartype/_util/utilcallable.py
```diff
@@ -19,39 +19,68 @@
...
+            filename = "<builtin>"
+        # Check the function/method contains a code object with the declaring filename.
+        elif (getattr(func, "__code__", None) and
+              getattr(func.__code__, "co_filename", None)):
+            filename = os.path.abspath(func.__code__.co_filename)
+        # Callable was dynamically declared.
+        else:
+            filename = "<string>"
+
+    return filename
```
leycec, line 86: **Superb first draft.** Because the whole point of this exercise that has depleted your essential vitality is absurd efficiency, I have comments that will make you begin to dread life itself:

* **Object attribute lookup is surprisingly expensive,** because it implicitly entails a dictionary lookup and thus call to the `dict.__getitem__()` method. This means *every* `.` operator (e.g., `sys.modules`, `func.__module__`). In Python microoptimization, we like to optimize away repeated object attribute lookup by only performing the lookup once. And by we, I mean me. I am horrible! *Fear my disgusting ear whiskers.*
* The prior point means that it's *much* more efficient to import attributes directly (e.g., `from sys import modules`) rather than access attributes through modules imported directly (e.g., `sys.modules`).
* I'm *pretty* sure (but *not* entirely sure) that all pure-Python callables have code objects. Equivalently, if a function or method isn't C-based, it really had better have a `__code__` attribute. Ergo, `getattr(func, "__code__", None)` *should* be safely reducible to just `func.__code__`.
* **Bound methods wrap unbound functions.** Fortunately, distinguishing between bound methods and unbound functions is trivial and efficient. If it's an instance of `beartype.cave.MethodBoundInstanceOrClassType`, it's a bound method; if it's an instance of `beartype.cave.FunctionType`, it's an unbound function. You are probably thinking: "Will he stop talking?" No! No, I won't. I like talking. This distinction matters, because **only unbound functions have `__code__` attributes.** So how do you get the `__code__` attribute for a bound method? Simple: you get the unbound function wrapped by that bound method first via the `__func__` attribute (e.g., `func.__func__.__code__` for bound methods). *Python is Hell.*
* For efficiency, simplicity, and safety, it's often best to initialize the local variable you *eventually* want to return to a safe fallback first. In this case, that means initializing `filename = '<string>'` **as the first statement of this function.** If you do that, you'll hopefully find that the resulting implementation becomes much simpler, more readable, and possibly even faster. These are all good things! Of course, I might also be wrong and doing that will just destroy the structure of the entire fragile edifice. @leycec bestows "wisdom" even as he rubs his fearsome bushy eyebrows.
* Whenever you see repeated code fragments (e.g., `sys.modules.get(func.__module__)`), you just know you've violated [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). That's bad, because it's slower than it otherwise could be. Thankfully, we've now initialized `filename = '<string>'` beforehand, which means that whole conditional on line 67 can be unravelled a bit.
* [The `__file__` attribute of *all* modules excluding one edge case that we don't care about is guaranteed to be absolute rather than relative under Python >= 3.4](https://stackoverflow.com/a/22866630/2809027). That means we can drop the first `os.path.abspath()` call, which would just reduce to a noop in 99.99999% of all calls anyway. *Down with the noop!*
* [The same applies to the `co_filename` attribute of code objects under Python >= 3.9](https://bugs.python.org/issue20443). I'm inclined to ignore relative `co_filename` attributes under older Python versions, as (again) this is only needed for non-mission-critical error messages. *Down with Python < 3.9!*
* We can be even cleverer with our `getattr()` calls and pass `filename` rather than `None` as their third parameters, **which makes the code stupidly unreadable but deliciously fast.** I choose deliciously fast. Let someone else maintain this.

All in all, I'm just another thick brick in the wall:

```
    filename = '<string>'

    # Grab location of the module declaring class
    if isinstance(func, type):
        # Check the Class has a module thats registered with the system.
        func_module_name = getattr(func, '__module__', None)

        if func_module_name is not None:
            func_module = modules.get(func_module_name)
            filename = getattr(func_module, '__file__', filename)
    else:
        # Built-in types are callables with no .py source file.
        if isinstance(func, CallableCTypes):
            filename = '<builtin>'
        # Check the function/method contains a code object with the declaring filename.
        else:
            if isinstance(func, MethodBoundInstanceOrClassType):
                func = func.__func__

            filename = getattr(func.__code__, "co_filename", filename)
            
    return filename
```

**Bwahah!** It's as beautiful as a plastic bag fluttering in the wind that reminds me of the ephemerality of my own decaying meat suit. I am crying. :crying_cat_face:

leycec (commented):
Welp, that was awesome. Let us now share a steaming mug of seventy year-old gunpowder green tea as we mutually soak in all the glorious code.

beartype/_util/utilcallable.py
```diff
@@ -19,39 +19,68 @@
...
+    See Also
+    ----------
+    :func:`inspect.getsourcefile`
+        Inefficient stdlib function strongly inspiring this implementation,
+        which has been highly optimized for use by the performance-sensitive
+        :func:`beartype.beartype` decorator.
+    '''
+    if callable(func) is False:
+        raise _BeartypeUtilCallableException(
+            f"Attempted to pass non-callable:{repr(func)}")
```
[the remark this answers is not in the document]
Heliotrop3: I concur with this ideal - It ha been (somewhat) promptly corrected in the following commit.

beartype/_util/utilcallable.py
Heliotrop3: >  I'm _pretty_ sure (but _not_ entirely sure) that all pure-Python callables have code objects. Equivalently, if a function or method isn't C-based, it really had better have a `__code__` attribute. Ergo, `getattr(func, "__code__", None)` _should_ be safely reducible to just `func.__code__`.
> 
>  Likewise, _all_ code objects should have `co_filename` attributes. For in-memory callables, the value of that attribute should _already_ be `<string>`. Ergo, `getattr(func.__code__, "co_filename", None)` should also be safely reducible to just `func.__code__.co_filename`.
> 
>  Likewise, _all_ classes should have `__module__` attributes. For in-memory classes, the value of that attribute should _already_ be `None`. Ergo, `getattr(func, "__module__", None)` should also be safely reducible to just `func.__module__`.

>  [The `__file__` attribute of _all_ modules excluding one edge case that we don't care about is guaranteed to be absolute rather than relative under Python >= 3.4](https://stackoverflow.com/a/22866630/2809027). That means we can drop the first `os.path.abspath()` call, which would just reduce to a noop in 99.99999% of all calls anyway. _Down with the noop!_
> 
>  [The same applies to the `co_filename` attribute of code objects under Python >= 3.9](https://bugs.python.org/issue20443). I'm inclined to ignore relative `co_filename` attributes under older Python versions, as (again) this is only needed for non-mission-critical error messages. _Down with Python < 3.9!_

I assumed such, but wasn't sure about bad actors. I'm ever increasing my <s>power 😈 </s> knowledge of OOP and am not aware if someone could, through omission or some Pythonic anti-magic, effectively delete a dunder. (i.e. The dunder difference between Mapping v. MutableMapping) 


>  Sadly, the prior point suggests that tests aren't _quite_ there yet. Tests shouldn't have passed, but they did. I get tired just thinking about tests, so I'll leave this in your ingenious hands. I apologize for everything.

Yeah, I'm still getting the hang of the testing bit. In this particular instance writing the test cases before developing the solution proved, and is proving, to be [challenging to understand](https://youtu.be/3VA9vuId3ds?t=16)

Clean is good! My initial implementation felt a little chunky. I did make one change based on my own interest in Byte-Code/Assembly. It's an odd fascination whose practicality in Python is questionable at best but here we are nonetheless 
At a byte-code level in ***Python 3.9*** this call is roughly 2 instructions longer  than it's equivalent. Why? - the  `is not false` convention uses the `IS_OP` code which requires loading in the value of  `None` to check whether the current value in fact the opposite.
>```
>         if func_module_name is not None:
>             filename = getattr(modules[func_module_name], '__file__', filename)
>```
```
 79     >>   88 LOAD_GLOBAL              3 (isinstance)
             90 LOAD_FAST                0 (func)
             92 LOAD_GLOBAL              9 (MethodBoundInstanceOrClassType)
             94 CALL_FUNCTION            2
             96 LOAD_CONST               6 (None)
             98 IS_OP                    1
            100 POP_JUMP_IF_FALSE      108

```
Contrast that with the piece of code which will throw `False` in the case the value is `None`
```
if func_module_name:
    filename = ....
```

```
 79     >>   88 LOAD_GLOBAL              3 (isinstance)
             90 LOAD_FAST                0 (func)
             92 LOAD_GLOBAL              9 (MethodBoundInstanceOrClassType)
             94 CALL_FUNCTION            2
             96 POP_JUMP_IF_FALSE      104
```
Note: I am not sure what this looks like in a pre-3.9 environment

leycec: > At a byte-code level in Python 3.9 this call is roughly 2 instructions longer than it's equivalent.

**Oh, shit.** We got a bytecode badass here. It's awesome and horrifying that you went deeper than even I – the self-proclaimed neckbeard on the frozen lake with a bald head, two cats, and no hockey sticks – have gone on microoptimization. And I go deep.

I'm depressed that `func_module_name is not None` is slow. It makes sense, but that doesn't make me any less depressed. This is why we defer to the bytecode badass. I will have further need of your special "services" when we go deep into the bytecode hole with `beartypecache`, a prospective project that will involve cray-cray Abstract Syntax Tree (AST) and bytecode manipulation destined to destroy this timeline. See `codemain.py` if sanity-eroding immolation-by-code-complexity is your thing.

leycec:
Someone get the fire extinguisher! We've got an explosion of awesome code here! :fire_extinguisher: 

Thanks *soooo* much for all the intrepid investigations into the deep end that is Black-magic Python, where none but only the most steadfast of bytecode loyalists fear to tread. You are such a man, Tyler.

PR #23, opened 2021-02-10
refactor: Implement threading.Lock

Heliotrop3:
The **only** reason this commit is marked as refactor is:
- Change from 'RLock' to 'Lock'
- Invert logical check in `__contains__`

Arguably the smallest change was threading.Lock. Removing the R from
'RLock' seems suspiciously easy... However, as per the FIXME, this
appears to be how `utilcacheall` implemented thread safety. Likewise
several articles seemed to indicate this is the case so here we are.
Although notably this **should** resolve the FIXME from `utilcachelru`.

Everything else is geared around putting a shine on the bear's coat by
tuning the formatting of comments and doc strings. Which, in that respect,
the latter item would fall under this category as well. This was inspired
by the recent commits including commentary around potentially templating
the lru strong cache. In short, making sure the bear contains the right amount
of noise to assist future development efforts in building off the preexisting
internal mechanism.

Signed-off-by: Tyler Huffman <43817374+Heliotrop3@users.noreply.github.com>


Heliotrop3:
Seems I broke one of the test cases... Obligatory 'It worked on my local machine' 😆 😢 .  To hopefully not burn CI/CL minutes, I'm going to close and resubmit after I figure out which part broke between testing and touching up the doc strings.

leycec:
Infinite busy wait, Y U do this 2 us!?!?

In related CI news, this PR consumed [**6h 0m 39s** or **360 minutes** of scarce CI time](https://github.com/beartype/beartype/actions/runs/553876638). That's... not good. But that's not *your* fault. Well, I mean, it sort of is. But that's mostly *our* fault for failing to cap `tox` execution time to some sane maximum. I may have no practical idea how to actually do that, but at least I now have incentive.

Also, you meant well and that's what counts. Refactoring away slow `RLock` usage into fast `Lock` usage is undeniably clever, inscrutably brutal, and deserving of raucous pom-pom applause from scantily clad cheerleaders – who are sadly in short supply thanks to social distancing regulations in post-2020 emoji. :dancing_women:

Before submitting your next PR Harbinger of Computational Doom, **please ensure I've committed something forcing `tox` to timeout first.** Thankfully, [`tox` supports a medley of timeout options](https://tox.readthedocs.io/en/latest/_modules/tox/config.html), so this mostly looks trivial.

Unrelatedly but pertinently:

* As you doubtless already know because you have achieved the pinnacle of software achievement, `LRUCacheStrong` *is* currently thread-safe. I just forgot to remove the preceding `FIXME:` comment after making it thread-safe. This is why no documentation is better than bad documentation. Speaking of which...
* **Please preserve relevant docstrings and comments,** if you can. I know, I'm bloody verbose and this rankles like an inopportune leech in all the wrong places. But my future self is fond of my past self being bloody verbose, because my future self is bloody forgetful and can never remember how, why, when, or what my past self was thinking when he wrote that bloody awful code – and then my future self inevitably thinks he's clever and rewrites the uncommented code written by his bloody stupid past self only to belatedly realize after twelve failing unit tests that his bloody stupid past self was actually on to something and had gotten it right the first time but failed to document anything and here we are. *This is why we can't have terse things, folks.*

Thanks again for all the savoury code morsels, Tyler! Let's gut that `RLock` like a fish.

leycec:
**Hurrah!** We communally rejoice as our continuous integration (CI) workflow now caps tests to **5 minutes.** This is almost certainly not enough, which is bad. But I'm curious whether it really works, so let's let it randomly fail on a subsequent push or PR and see what happens. *It is good.*

Issue #25, opened 2021-02-16
PEP 561 compliance


harens:
Hi @leycec and the @beartype organisation 👋 

I'm currently rewriting large parts of a [project of mine](https://github.com/harens/seaport), and I'm planning on using beartype 🐻 with it! (because beartype is _awesome_ 😎)

Part of the rewrite involves creating a python module that users will be able to import. Currently, my project is [PEP 561 compatible](https://www.python.org/dev/peps/pep-0561). However, beartype causes some issues with this making it no longer the case.

For instance, if I were to run `mypy --strict` on my code, I now get lots of the following errors:

```
Untyped decorator makes function "..." untyped
Skipping analyzing 'beartype.cave': found module but no type hints or library stubs
Skipping analyzing 'beartype': found module but no type hints or library stubs
```

This then has the knock on effect that the functions in my project appear untyped when imported, making it no longer PEP 561 compliant.

Assuming beartype is properly annotated, which I'm  200% sure it is, it should hopefully be as simple as adding a `py.typed` file in the [project directory](https://github.com/beartype/beartype/tree/main/beartype). See the [mypy docs](https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-type-hints-for-third-party-library) for more info. This could then be another PEP ticked off in the [compliance](https://github.com/beartype/beartype#compliance) list.

Since you have a lot (and I really do mean a lot) more experience with these different PEPs and type checking than I do, please let me know if I'm missing anything. Thank you for your help and for maintaining this great project.

leycec:
> Hi @leycec and the @beartype organisation :wave:

**Oh! Oh!** It's Haren, everyone's favourite macOS package maintainer, GitHub infrastructure expert, and just all-around inspiring code hero – back from early retirement with a hard-hitting deep dive into PEP compliance that will make us all better, stronger, and more employable Pythonistas. *It's gud.* :pleading_face: 

> I'm currently rewriting large parts of a project of mine, and I'm planning on using beartype :bear: with it! (because beartype is awesome :sunglasses:)

Oh, you. Sweet pillow talk and strenuous promises of a new reverse dependency on `beartype` guarantee zero-day issue turnaround here at @beartype, Ltd. (Choose `beartype`: because bears would do it, too.)

> However, beartype causes some issues with this making it no longer the case.

![Noooooo!](https://i.kym-cdn.com/entries/icons/original/000/000/854/vader_NOOOO.jpg)

[...]

PR #26, opened 2021-02-17
PEP 561 x 2: Electric Boogaloo 🎸

harens:
So! Here we are from the depths of https://github.com/beartype/beartype/issues/25.

On my machine, this has brought the __number of mypy errors from ~155 to ~98__ (we do not speak of `mypy --strict`). For your machine, it might be slightly more or slightly less depending on virtual environment/deps installed/etc.

Just to note, that this does not fix the errors a user will face from https://github.com/beartype/beartype/issues/25#issuecomment-780618689, at least from my local testing it doesn't anyway. The purpose of this PR is to hopefully act as another stepping stone on our journey to becoming PEP 561 compliant, and so make it more likely to fix the errors that occur.

Another thing, although your codebase is extremely well documented, it is also very scary 😨 (only supreme warlock @leycec ⚔️ and master wizard @Heliotrop3 🧙 know how to traverse it). Hopefully I haven't got in the way of any of your speed optimisations or anything else. I think the only potential issue is `IS_PYTHON_AT_LEAST_3_X`, which I almost certainly should've used to deal with `typing`/`types` imports...

_P.S. Don't worry, I put a more appropriate commit title :)_


beartype/_cave/abc.py
```diff
@@ -41,6 +41,7 @@
 
 # ....................{ IMPORTS                           }....................
 from abc import ABCMeta, abstractmethod
+from typing import Union
```
Heliotrop3, line 44: @leycec, I'll take this opportunity to raise a point about the future state of the `Union` type  - As per [PEP 604](https://www.python.org/dev/peps/pep-0604/), in Python 3.10 half of a C or statement - `|` - essentially aliases the Union type. I.E.:

```
def bear_func(param1: str, param2: int) -> Union[str, int]:....
```
will be equivalent to:
```
def bear_func(param1: str, param2: int) -> str | int: ....
```
Most of the work I've done has been on the data structure and testing side so I'm less familiar with the actual type hinting implementation. In short, I need to research the aforementioned implementation before opening a ticket on the matter. Barring unforeseen events, the plan was to tackle this over the weekend along with #24.

leycec:
```
def bear_func(param1: str, param2: int) -> str | int: ....
```

Freyja above, but that's *really* smokin' hot. If there's one thing we need, it's money and ~~drugs~~ vaccines. If there's two things we need, it's more set, group, and category theory operators pervading every aspect of our all-too-brief lives as shambolic meat suits on this pale blue dot.

Thanks as always for the apropos heads up, Master Tyler. I had no idea PEP 604 was even a thing, but now I do and I'm a better Canadian man for it.

leycec:
**OMFG.** This is surely the Work That Transforms a Lack of Codebase Quality into Quality (WTTALOCQIQ)! Also, this would ordinarily be the place where I waste everyone's time and lifeforce with a pedantic line-by-line code review which only exhibits my own appalling ignorance and shameless tendency toward self-aggrandizement.

But I think we can forego that here. :wink: 

> On my machine, this has brought the number of mypy errors from **~155** to **~98**

:bangbang: 

> (we do not speak of `mypy --strict`).

`mypy --strict`, *you are dead to me.*

> Just to note, that this does not fix the errors a user will face from #25 (comment), at least from my local testing it doesn't anyway.

**Urgh!** Despite our best efforts, utopia remains but a distant dream. Exactly as you say in #25, the last remaining hurdle is my own intransigence. Specifically, this choice nugget in `beartype.__init__`:

```
# Intentionally defined last, as nobody wants to stumble into a full-bore rant
# first thing in the morning.
__all__ = ['STAR_IMPORTS_CONSIDERED_HARMFUL']
```

`STAR_IMPORTS_CONSIDERED_HARMFUL`, we will miss you. `</sigh>`

Let's try either removing that *or* reducing that to something sane: e.g.,

```
__all__ = ['beartype']
```

It is tragic, but I can't seem to replicate the `error: Module 'beartype' does not explicitly export attribute 'beartype'; implicit reexport disabled` error on my end – even with `mypy --strict`. Is this a mypy 0.800 thing? Gentoo still has me shackled to mypy 0.790, which it claims without evidence to be stabler than 0.800. This is why we can't have good quality.

While it *will* pain me deep inside to see `STAR_IMPORTS_CONSIDERED_HARMFUL` go, I think I speak for us all when I say that conformance with community standards is a wee more important than muh idiosyncratic feels.

> Another thing, although your codebase is extremely well documented...

Aww, shucks and golly gee whiz willikers. :relaxed: 

> ...it is also very scary :fearful:

`(╯°Д°)╯︵ /(.□ . \)`

> I think the only potential issue is `IS_PYTHON_AT_LEAST_3_X`, which I almost certainly should've used to deal with `typing`/`types` imports...

Yes! Yes, you should've! Except... **tests pass.** So, no? No, you shouldn't have?

I defer to tests in all things in the mistaken belief that they're actually covering all edge cases. If tests say you're a golden flying baby, you're a golden flying baby.

leycec:
**It is brilliant.** It is in! Welcome to the wilfully dysfunctional but jovial @beartype family. *One of us, one of us.*

PR #27, opened 2021-02-22
test: Update test cases for LRU Strong Cache

Heliotrop3:
This commit preps the code base for a larger commit of a thread safe LRU strong cache using the simple `threading.Lock()` method.

This commit specifically:
 - Updates the expected errors raised on failure to built-in exceptions.
 - Revises comments in the test cases for clarity
 - Redefines key,value pairs for ease of debugging


Heliotrop3:
Not 100% sure how to go about this one: submitting the large request will fail as it depends on an updated test file. Yet, the larger change itself needs to be put through the scrutinizer. 

 Without pushing the larger commit to this branch, I figure there are a couple of items that we can scrutinize.

>  - Updates the expected errors raised on failure to built-in exceptions.

The reason for this was two-fold:
 - Consistency
 - Clarity

The localized parent methods employed by LRU strong cache implicitly raises a couple of errors:
 - TypeError if the key isn't hashable
 - KeyError if `__getitem__` doesn't find the key


We could change this behavior across all of the cache but that would involve some try and except blocks where we're catching the implicit exception and raising our own in its stead.  This isn't difficult but would slightly bloat the LOC and bytecode.
The other aspect - clarity - is the built-in exceptions capture and convey the error in as streamlined a manner as possible.

For example, below is the only effected snippet from the LRU cache ***before*** the exception change.
```
        if not isinstance(size, int):
            raise _BeartypeUtilLRUCacheException(
                f'LRU cache capacity {repr(size)} not integer.')
        elif size <= 0:
            raise _BeartypeUtilLRUCacheException(
                f'LRU cache capacity {size} not positive.')
```
And here is said snippet ***after*** the change.
```
        if not isinstance(size, int):
            raise TypeError(
                f'LRU cache capacity {repr(size)} not integer.')
        elif size < 1:
            raise ValueError(
                f'LRU cache capacity {size} not positive.')
```
I propose the second snippet contains less noise with just as much information.

> - Revises comments in the test cases for clarity

This was more for my sake - I wanted to track what test cases I had covered so I reworded the comments to explicitly state which methods and cases where being tracked were it wasn't obvious.

> - Redefines key,value pairs for ease of debugging

This one, it pained me. I like easter eggs as much as the easter kangaroo but the `threading.Lock()` implementation was not without a few trips to the PDB. I found swapping the values to bland and boring key,value ... well, values, reduced the amount of squinting I needed to do.

leycec:
> This commit preps the code base for a larger commit of a thread safe LRU strong cache using the simple `threading.Lock()` method.

**Brilliant!** I approve of forward-thinking audacity in both `beartype` and avant-garde graphic novels penned by obscure British artists.

> Revises comments in the test cases for clarity

**Sweet!** Code Clarity are my two middle names. Let this be so.

> Redefines key,value pairs for ease of debugging

...you do realize you just censored T.S. Eliot's [*The Love Song of J. Alfred Prufrock*](https://www.poetryfoundation.org/poetrymagazine/poems/44212/the-love-song-of-j-alfred-prufrock) – only the greatest long-form epos in any human language, that's all – don't you?

I mean, I applaud human-readability, -debuggability, and -usability as much as the next die-hard Vim fundamentalist. I *do.* Really. But you'll never make me truly like any of those things.

...oh, very well. Let's just censor T.S. Eliot then. Why not? Goodbye, carefree witticisms. Goodbye. This is how the codebase dies: not with a bang, but a pull request.

I begrudgingly accede to this request, but only because you have a valid point. Even I wouldn't want to PDB that hot mess! `</wheezy_sigh>`

> Updates the expected errors raised on failure to built-in exceptions.

**Oh, no.** Nope! Nadda! Nil! Nihilism! And other "N"-prefixed synonyms of those things!

In general, explicitly raising builtin exception types from third-party dependencies is bad. Why? Ambiguity. Whereas builtin exceptions are commonly raised by both CPython and the stdlib, `_BeartypeUtilLRUCacheException` is private and should thus *never* be raised by anyone – especially including us. Everyone expects to see `TypeError` and `ValueError` exceptions, but no one expects to see `_BeartypeUtilLRUCacheException`.

We need to preserve that distinction. If a user receives an unhandled `TypeError` or `ValueError`, they'll correctly assume that to be due to something *they* did with respect to either CPython or the stdlib. Us raising those exception types invalidates that correct assumption. If someone sees a wacky private exception type like `_Beartype{insert_something_crazy_here}`, the first thing we want them to think is:

> **"Yup. That's a bug, and it sure ain't mine. Better report that to the @beartype issue tracker stat before someone loses a gangrene-infested leg over this!"**

No one expects to see builtin exceptions from third-party dependencies. The only exceptions I know of are:

* Raising `KeyError` on missing container keys.
* Raising `ImportError` on unimportable modules.
* Raiseng `TypeError` on unhashable objects.

In all three cases, **the builtin exception message is preserved exactly as is** (e.g., `TypeError: unhashable type: '{classname}'`). This enables users to correctly detect that those errors correspond to what everyone assumes those errors correspond to.

In our case, we need `beartype`-specific exception messages for `beartype`-specific exceptions. Raising builtin exception types instead would prevent users from catching and handling `beartype`-specific exceptions, **which would be horrible.** Specifically, no one could do this:

```
from beartype.roar import BeartypeException
try: 
    # Do something dark, nasty, and unmentionable with @beartype,
    # probably in a test suite manned by the NSA somewhere.
    @beartype
    def break_the_bear(muhuhahahahahahahahahahahahahahahahaha: int): pass
except BeartypeException:
    print('Checkmate, beartype. Checkmate.')
```

See also [this pertinent StackOverflow answer](https://stackoverflow.com/a/2180803/2809027), which I will now shamelessly plagiarize:

> The main problem with doing it that way is that clients of your code have a hard time being precise in their exception handling; in order to catch that specific exception, they would have to do string matching on the exception object they caught, which is obviously fragile and tedious.

Heliotrop3:
> 
> > Updates the expected errors raised on failure to built-in exceptions.
> 
> **Oh, no.** Nope! Nadda! Nil! Nihilism! And other "N"-prefixed synonyms of those things!
> 
> In general, explicitly raising builtin exception types from third-party dependencies is bad. Why? Ambiguity. Whereas builtin exceptions are commonly raised by both CPython and the stdlib, `_BeartypeUtilLRUCacheException` is private and should thus _never_ be raised by anyone – especially including us. Everyone expects to see `TypeError` and `ValueError` exceptions, but no one expects to see `_BeartypeUtilLRUCacheException`.


Hmmm, I'm going to try to eat this elephant ear of understanding one bite at a time. Our aim is the same - clear concise beartype goodness - but our points run counter to one another which only peaks my already unhealthy curiosity. Specifically, I'm making the claim the built-in exception provides clarity which you're refuting stating in fact it provides ambiguity.

I'd like to increase my understanding by gaining clarity on how being a private module effects the error messages. If I'm understanding your rationale correctly, essentially we need a private exception to give the user the ["It's not you it's me"](https://www.youtube.com/watch?v=2uAj4wBIU-8) should the LRUCache ever raise an error in the wild. This makes sense - we **do** want to indicate when something [is not their fault](https://youtu.be/ZQht2yOX9Js?t=177).  Likewise if my understanding is correct, this also provides the ability for granular error handling. Finally, the StackOverflow article rings true - text matching an error message is tedious, fragile, and adds maintenance overhead.

I think these are all convincing points - so much so I'm inclined to change my position and agree with you're reaction:
>Oh, no. Nope! Nadda! Nil! Nihilism! And other "N"-prefixed synonyms of those things!

This does, however raise a question about how implicitly raised errors are handled. Specifically, every exception raised by LRUCache - except for the ValueError on initialization - is either a TypeError or a KeyError. Of those exceptions, in my implementation all but three are called implicitly via the localized methods of the parent class:  two on initialization and one in the `__get__` because I did a thing for DRY. 

 - Will those implicit raises reflect the private error or will we need to structure the code differently to raise the exceptions through our private exception?

<br>

I based this experiment on a nearly decade old PyCon talk [Stop Writing Classes](https://www.youtube.com/watch?v=o9pEzgHorH0). While I think the first roughly two minutes is generally solid logic - customers use software for features not amount of code therefore if you can get more for less do it - I think the speaker struggled in the Q&A portion.  Ultimately I see this as YAOG (Yet Another Opportunity for Growth)... the acronym is a work in progress 😆.

leycec:
> I'm going to try to eat this elephant ear of understanding one bite at a time.

:elephant: :ear: :ear_of_rice: :yum: 

> "It's not you it's me"

You're giving me the "It's you, not me" routine? I invented the "It's you, not me" routine. Nobody tells me "It's them, not me." If anybody, it's me!

> This does, however raise a question about how implicitly raised errors are handled.

I have now donned the ceremonial black hat of venerable knowledge, accumulated wisdom, and quick and dirty cheap thrills. :tophat: 

> ...a `KeyError` in the `__get__` because I did a thing for DRY.

This is wise and just, for DRY is the ancient enemy we do continual battle with. My scant and skimpy advice here would be to explicitly mimic the exact `KeyError` exception format raised by the `dict` superclass on missing keys.

Yes, it's `f'KeyError: {missing_key}'`, which is *hilariously* non-human-readable, but it's best not to violate de-facto standards, even if we never expect these exceptions to ever percolate up the call chain to users, which we don't.

I give you this **Magical Sword of Software Engineering +1** that you may deftly slay the goblins still gibbering in the dark corners of the codebase. :hocho: 

> Will those implicit raises reflect the private error or will we need to structure the code differently to raise the exceptions through our private exception?

I'm not sure I understand the question, because my last reservoirs of intelligence are like the fleeting pirate ghost ship that dissipates by dawn – and it's almost dawn. Curse the wretched light! :city_sunrise: 

Implicit raises are perfectly fine and indeed preferable when you're masquerading as something you're not – in this case, an LRU cache masquerading as a dictionary. It's all about preserving rather than violating expectations, really. Everyone expects the `__getitem__()` method for a dictionary to raise well-formatted `KeyError` exceptions, so we preserve that expectation by either deferring to the superclass `dict.__getitem__()` method where feasible or just mimicking that method's behaviour by explicitly raising `KeyError` exceptions of the same format otherwise.

We usually raise private exceptions everywhere else. No one expects the `__init__()` method for a dictionary to raise `ValueError` exceptions, for example, because the standard `dict` type doesn't do that. So we raise private exceptions instead, which hopefully never percolate up the call chain but which, should they ever do so, will notify everyone that we've really cocked it up badly and should now pay with our online pride and a badge of shame shackled to the necks of our GitHub avatars like a stinking dead albatross in a Victorian-era poem about seamen or something.

> I based this experiment on a nearly decade old PyCon talk Stop Writing Classes.

**Yeah.** That's an awesome PyCon talk right there as demonstrated by the fact that we're still talking about that talk nine years later. I don't disagree with much here, because I am sane and this is a sane talk. I only disagree with this capitalist-centric sentiment:

> *"I hate code and I want as little of it as possible in our product"*
> *— me, always*

This is a fascinating train of thought that I myself have never had.

I love code! I love it so much I do it for free! It's one of the few aspects of physical reality that actually makes sense! Did I mention I *really* love code?

Conversely, I hate that the standard dilemma in our industry is to either:

1. Get paid giant piles of money-stuffed laundry bags writing closed-source code that no one can ever see.
1. Get paid only in sympathy, pity, and shared tears writing open-source code that everyone can see and eventually has to when your code fails during their eleventh-hour Friday night crunch.

> ...if you can simplify the codebase do it.

If I obeyed this maxim, `beartype` wouldn't exist. Of course, this *is* a useful maxim under standard operating for-profit procedures where the codebase is merely the means to the ends which is money.

We have different ends here. Our ends are to be the best type checker in any dynamically-typed language where best means fastest, most compliant, most accurate, and least opinionated (in that order). As I subjectively see it, the main lesson of `beartype` is that **being the best as measured by those metrics means writing a lot of code** – like, a mammoth pile of code so grotesque in its immensity that even `wc -l` sputters and chokes on its Canadian-brand cigarillo when you point it at our codebase.

"Fast" is the canonical example. It's *really* easy to write slow Python. By default, Python is slow. It's one of the slowest languages to run, fastest languages to write, and easiest languages to maintain code for that we have yet invented as a feces-flinging species of bipedal primates, which is something remarkable and depressing. If we didn't care about speed, we'd be `typeguard`, which reduces to a single reasonably-sized module.

Instead, we care about speed, which means we wrote **hundreds of thousands of excess code** we otherwise didn't need to. Because extracting speed from the sluggish single-core boa constrictor that is CPython takes an obsessive-compulsive anal fixation on microoptimization, caching, and DRY violations. I mean, we're so single-mindedly obsessed with speed here [we wrote not just *one* but *two* `O(1)` runtime type checkers in a two-pass Gordian knot of irreducible complexity](https://github.com/beartype/beartype#id51):

1. A ludicrously fast non-recursive `O(1)` runtime type checker that just tests types *without* raising human-readable exceptions.
1. A ludicrously slow recursive `O(1)` runtime type checker that both tests types *and* raises human-readable exceptions – which turns out to be much, much harder, slower, and less fun overall than just testing types.

So, code. Stinking piles and steaming heaps and veritable mountains of extraneous code – all to do something everyone else has already done, just excruciatingly fast, which surprisingly no one has ever done, probably because something best describable as "excruciating" is the opposite of low-hanging and fun for most people.

We are not most people. Welcome to `beartype`. :wink:

leycec:
**Wunderbar!** I don't mind admitting that your deep macro-optimizations exude brilliance from every statement, branch, and call. I also don't mind admitting that I should have merged this as is a week ago. So, why didn't I do that rational and decent thing?

Because `beartype` got repeatedly sledgehammered in its fat paunch while peacefully hibernating in its winter den. Specifically, [a professional test engineer publicly disclosed that `beartype 0.5.1` raised critical exceptions on a certain kind of PEP-compliant type hint referred to only as a "subscripted generic,"](https://github.com/beartype/beartype/issues/29) which you do *not* want anything to do with. <sup>Stay clean. Stay safe.</sup>

The bear immediately panicked, rent the well-meaning intruder limb from limb, spent a harebrained hair-pulling week fixing the problem, released `beartype 0.6.0`, and promptly went back to sleep matted in blood, brains, and viscera. Things got sufficiently crazy that I lost even more sanity points and hair follicles. I honestly *cannot* believe I didn't break your PR in the process. By a late-winter miracle I can only attest to Freya Herself, I didn't.

**Let's merge this before I break this.** Thanks for the stupendous outpouring of open-source volunteerism, @Heliotrop3. If you'd ever like me to write a Python-specific job recommendation for you, please do let me know; I promise to be cordial, polite, and terse, which is to say I promise to pretend *not* to be me.

## Informal Code Review, Because We Care

Lastly, a trivial code commentary. You're so smart it burns my eyes. Clearly, you love DRY. I respect that. I like DRY – but I don't quite *love* DRY. Why? Because this is software engineering, there's a tradeoff here between violating DRY and optimizing caching. In this case, you avoid violating DRY by defining a common private `_push()` utility method called by all three higher-level dunder methods (i.e., `__getitem__()`, `__setitem__()`, and `__contains__()`). That's good, right?

Well, it's certainly *clever* – *brilliant,* even. But it's sadly suboptimal. That doesn't mean that it's *bad,* per say. It's certainly not *bad.* After all, it's brilliant! By definition, that which is brilliant cannot be bad. Right?

But it's still sadly suboptimal, because **[function and method calls are outrageously expensive in Python even by Guido's lax standards](https://stackoverflow.com/questions/22893139/why-is-a-function-method-call-in-python-expensive).** In this case, you avoid violating DRY by forcing an additional stack frame onto the call stack – which is sufficiently slow that it kinda violates the whole purpose of caching, which is to be stupidly fast. Right?

The solution is to **foolishly copy-and-paste** the body of the `_push()` method into the three higher-level dunder methods calling that method. Note that doing so *also* enables us to microoptimize away the conditional test `if value is None:` in the case of the `__setitem__()` method, except...

## Informal Code Review Exposes Bad Stuff

**Woopsie.** It looks like your logic contains a steaming dragon that is breathing fire all over the hapless villagers – *and nobody even knew about it.*

What happens when someone explicitly sets a dictionary key first to a non-`None` value and then to `None`? Under your refactored implementation, the answer is the dictionary summarily defecates all over itself by silently resetting that key back to its initial non-`None` value. *Oh, Gods.*

Behold, the Code Dragon Smaug in All His Hideous Smugness:

```
>>> smaug_cache = LRUCacheStrong(size=3)
>>> smaug_cache['Bwahaha!'] = 'Help! Someone!'
>>> smaug_cache['Bwahaha!'] = None
>>> smaug_cache['Bwahaha!']
'Help! Someone!'
```

I think Code Dragon Smaug just broke the Internet. There are two solutions, the easy one and then the one that we should actually do:

* [*the easy one*] Rewrite the signature of the `_push()` method to resemble:

```
from beartype._util.utilobject import SENTINEL
from typing import Optional
...
    def _push(self,
              key: Hashable,
              value: Optional[object] = SENTINEL,
              ...
    ) -> None:
```

* [*the one that we should actually do*] **Just inline away the `_push()` method as suggested above.**

Like, the `_push()` method is obviously fragile. It's clever, yes. But it's so clever that it introduced hard-to-debug issues that no one saw coming, also yes. Even if we apply the `SENTINEL` patch above, that *still* exposes us to the same issue if anyone ever tries to cache a `SENTINEL` value – which they will, because Murphy hates us and good luck ever debugging *that* one.

So we just guillotine `_push()` by inlining it away. By the way, did you know the French actually named the carts they used to haul the aristocracy to the guillotines with? I didn't. They were called [tumbrels](https://en.wikipedia.org/wiki/Tumbrel), apparently.

Let's gently tumbrel `_push()` to where it needs to go. :shopping_cart: 

## I AM SO SORRY

**Ugh!** I feel like such a bad dude. I rudely sat on this for a week and then puked all over everything. I should probably just merge this as I promised above, but... that `_push()` bug, bro. I just can't merge that. It's making both my eyes simultaneously twitch and water as I stare at its little squirming buggy feet.

**Ugh!** Take all my apologies. :weary:

Heliotrop3:
> 
> 
> **Wunderbar!** I don't mind admitting that your deep macro-optimizations exude brilliance from every statement, branch, and call. I also don't mind admitting that I should have merged this as is a week ago. So, why didn't I do that rational and decent thing?
> 
> Because `beartype` got repeatedly sledgehammered in its fat paunch while peacefully hibernating in its winter den. Specifically, [a professional test engineer publicly disclosed that `beartype 0.5.1` raised critical exceptions on a certain kind of PEP-compliant type hint referred to only as a "subscripted generic,"](https://github.com/beartype/beartype/issues/29) which you do _not_ want anything to do with. Stay clean. Stay safe.

No worries, It coincided with an unexpected family tragedy so in a bittersweet twist it worked out well.

> ## Informal Code Review, Because We Care

image
The more, the logarithmic merrier! 

> Lastly, a trivial code commentary. You're so smart it burns my eyes. Clearly, you love DRY. I respect that. I like DRY – but I don't quite _love_ DRY. Why? Because this is software engineering, there's a tradeoff here between violating DRY and optimizing caching. In this case, you avoid violating DRY by defining a common private `_push()` utility method called by all three higher-level dunder methods (i.e., `__getitem__()`, `__setitem__()`, and `__contains__()`). That's good, right?... [Sadly, `__push()` is] suboptimal, because **[function and method calls are outrageously expensive in Python even by Guido's lax standards](https://stackoverflow.com/questions/22893139/why-is-a-function-method-call-in-python-expensive).** In this case, you avoid violating DRY by forcing an additional stack frame onto the call stack – which is sufficiently slow that it kinda violates the whole purpose of caching, which is to be stupidly fast. Right? The solution is to **foolishly copy-and-paste** the body of the `_push()` method into the three higher-level dunder methods calling that method. Note that doing so _also_ enables us to microoptimize away the conditional test `if value is None:` in the case of the `__setitem__()` method, except...

I'm more or less ambivalent regarding DRY. I think its generally a good idea but there are cases where it won't be. 
I think the [post](https://stackoverflow.com/a/22895664) you linked captures the approach I defaulted to and in spectacular fashion proceeded to fail at:
>The way to produce efficient code is a posteriori, not a priori. First write the code so it is clean and maintainable, including function calls as you like. Then while it is running with a realistic workload, let it tell you what can be done to speed it up...

The problem was, past me seems to have not tempered their enthusiasm and heated the design kiln to such a degree we ended up making brittle the LRUStrongCache. Fortunately you spot checked the code and found the glaring mistake and was kind enough to bring the growth opportunity to my attention. What do I think the glaring mistake was? Ignoring the core backend of a LRU cache - resetting the keys order when they're accessed or mutated - should never change. <sup> We did say it was a glaring mistake :goberserk: </sup>Anyway, the result was something I thought was easy on the eyes which in reality - and hindsight - was kinda ugly and entirely off the mark. In short, I enthusiastically agree with your assessment:
> Let's gently tumbrel `_push()` to where it needs to go. 🛒

<br>

>I honestly cannot believe I didn't break your PR in the process.

Notably, the reason for this is the LRU cache is never used outside of beartype's tests - it's implemented but not currently used.

leycec:
> It coincided with an unexpected family tragedy...

I... I'm so sorry. 2020 and 2021 were rough years for humanity. When my father-in-law passed away two years ago from a rare variant of terminal myeloma, I learned the meaning of loss and grief and the necessity of empathy and compassion in all things.

My wife, myself, and our two plump cats hope the rest of 2021 plays out healthier and happier for you and yours. :sob: 

> Anyway, the result was something I thought was easy on the eyes which in reality - and hindsight - was kinda ugly and entirely off the mark.

It *was* easy on the eyes. It was super-compact and super-clever. *Too* clever. That was always my Achilles' Heel, too. When I'm too clever these days, I just know I've committed a terrible wrong and start checking my surroundings for midget assassins lurking in the gooseberry bushes.

Your refactoring is stupendous, of course. **It's in!**

If you'd like *yet another* compelling opportunity for stunningly dangerous growth, would you mind inspecting our internal commentary for a related caching data structure tentatively dubbed the `LRUDuffleCacheStrong`? If so, see line 887 of the `_pephint` submodule:

```
#FIXME: *WOOPS.* The "LRUCacheStrong" class is absolutely awesome and we'll
```

Thanks so much *yet again* for your tremendous contributions, Tyler. We'll all make it through 2021 [if it kills us](https://www.youtube.com/watch?v=T_qkVPZ8DJI).

PR #37, opened 2021-06-25
Use generic type signature for the decorator

### LINKED ISSUE #36: Beartype's Type signature breaks some features of pylance
Hovering over an `@beartype` annotated function in VSCode with the Python plugin (using Pylance) shows the function as

```
(method) foo: (*args: Unknown, **kwargs: Unknown) -> Unknown
```

This can be reproduced with a simple decorator function using the same type signature as `beartype`:

```
def beartype2(func: Callable) -> Callable:
    return func
```

jonathanmorley:
Should fix #36 🤞

### DIFF
--- a/beartype/_decor/main.py
+++ b/beartype/_decor/main.py
@@ -147,13 +147,15 @@ from beartype._decor._data import BeartypeData
 from beartype._util.cache.pool.utilcachepoolobjecttyped import (
     acquire_object_typed, release_object_typed)
 from beartype._util.func.utilfuncmake import make_func
-from typing import Callable, TYPE_CHECKING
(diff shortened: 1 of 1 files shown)


leycec:
**Fantastic.** Brilliant save, Sir Morley. I must confess here... I use a [battle-hardened CLI-oriented hand-built Vim](https://github.com/leycec/vimrc) as my IDE, which features about as much in-editor introspection as the Canadian bullfrog belching on a rotting lily pad outside our cottage. <sup>so not much</sup>

Thanks for both the detailed issue writeup at #36 *and* the working changeset. You the best Morley. :medal_sports:

Issue #38, opened 2021-06-29
No love for NotImplemented?


posita:
When overriding binary special methods (e.g., `__eq__`, etc.), returning [`NotImplemented`](https://docs.python.org/3/library/constants.html?highlight=notimplemented#NotImplemented) is a common convention, and generally omitted from returned types. But omitting it doesn't agree with poor old bear:

```python
from beartype import beartype

class Sneetch:
    @beartype
    def __init__(self, star: bool):
        self._upon_thars = star

    @beartype
    def __eq__(self, other) -> bool:
        if isinstance(other, Sneetch):
            return True  # Sylvester McMonkey McBean can pound sand
        else:
            return super().__eq__(other)

assert Sneetch(True) == Sneetch(False)  # all good
assert (Sneetch(True) == 42) is False  # boom
```

This will result in the following:

```python
beartype.roar.BeartypeCallHintPepReturnException: @beartyped Sneetch.__eq__() return "NotImplemented" violates type hint <class 'bool'>, as "NotImplemented" not bool.
```

Making `NotImplemented` an explicit part of the return type is apparently hard (see python/mypy#4791).

I'm not sure what the right play is here. One could probably make a pretty good case that one should omit the `@beartype` decorator and lean on the internals (in which case, this should be closed as not-a-bug). But I thought I'd raise it anyway to get another opinion.

leycec:
**Wow!** You're on a beartype issues roll today. Tell Matt what he's won, folks. It's... it's a Gold Star for Participation. Look, it's all the prize committee could afford. :medal_sports: 

> poor old bear

I chuckled. Then the bear growled and I stopped.

```
# Sylvester McMonkey McBean can pound sand
```

I chuckled even harder. Go pound sand, poor old bear.

*Phew!* I think I got that outta my system. So, your astute observations and keen intuitive sense for hot problem spots in my code are yet again spot on. Mypy (even moreso than official PEP standards) is the de-facto authority on everything typing-related. What mypy does, we do.

You're in luck... *again.* Mypy does [exactly what everyone wants](https://github.com/agronholm/typeguard/issues/178#issuecomment-800123199), which means I now need to ~~disembowel~~ refactor beartype to do exactly the same thing. Specifically, mypy avoids raising warnings or errors for binary dunder methods (and *only* binary dunder methods) both annotated by and returning `NotImplemented`. Unsurprisingly, that's exactly what we'll do too.

[...]

PR #40, opened 2021-07-04
Afford returning NotImplemented from all binary dunder methods

posita:
This expands on 57e5e99 to implicitly tolerate returning `NotImplemented` from binary dunder methods, irrespective of return type (not just those returning `bool`). Ref. #38.


posita:
Hmmm…I seem to have it a snag:

```
$ pytest -vv beartype_test/a00_unit/decor/code/test_codemypy.py  # <- same appears when run in tox
==================================================================== test session starts =====================================================================
platform darwin -- Python 3.9.5, pytest-6.2.4, py-1.10.0, pluggy-0.13.1 -- /Users/matt/.acct/.virtualenvs/dycelib/bin/python
cachedir: .pytest_cache
rootdir: /Users/matt/Documents/dev/dycelib/submodules/beartype, configfile: pytest.ini
plugins: gitignore-1.3
collected 1 item

beartype_test/a00_unit/decor/code/test_codemypy.py::test_decor_mypy_notimplemented FAILED                                                              [100%]

========================================================================== FAILURES ==========================================================================_______________________________________________________________ test_decor_mypy_notimplemented _______________________________________________________________
…
>       assert isinstance(the_seas + the_streams, TheCloud)

BeartypeCallHintPepReturnException = <class 'beartype.roar.BeartypeCallHintPepReturnException'>
TheCloud   = <class 'beartype_test.a00_unit.decor.code.test_codemypy.test_decor_mypy_notimplemented.<locals>.TheCloud'>
beartype   = <function beartype at 0x1099b6790>
the_seas   = <beartype_test.a00_unit.decor.code.test_codemypy.test_decor_mypy_notimplemented.<locals>.TheCloud object at 0x109ab1bb0>
the_streams = <beartype_test.a00_unit.decor.code.test_codemypy.test_decor_mypy_notimplemented.<locals>.TheCloud object at 0x109ab1b80>

beartype_test/a00_unit/decor/code/test_codemypy.py:123:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

__beartype_func = <function test_decor_mypy_notimplemented.<locals>.TheCloud.__add__ at 0x109ab6d30>
__beartype_raise_exception = <function raise_pep_call_exception at 0x109aa6c10>
__beartypistry = {'+6730761713772002360': (<class 'bool'>, <class 'NotImplementedType'>)}
args = (<beartype_test.a00_unit.decor.code.test_codemypy.test_decor_mypy_notimplemented.<locals>.TheCloud object at 0x109ab1b...eartype_test.a00_unit.decor.code.test_codemypy.test_decor_mypy_notimplemented.<locals>.TheCloud object at 0x109ab1b80>)
kwargs = {}, __beartype_pith_0 = <beartype_test.a00_unit.decor.code.test_codemypy.test_decor_mypy_notimplemented.<locals>.TheCloud object at 0x109ab17c0>

>   ???
E   NameError: name 'NotImplementedType' is not defined

__beartype_func = <function test_decor_mypy_notimplemented.<locals>.TheCloud.__add__ at 0x109ab6d30>
__beartype_pith_0 = <beartype_test.a00_unit.decor.code.test_codemypy.test_decor_mypy_notimplemented.<locals>.TheCloud object at 0x109ab17c0>
__beartype_raise_exception = <function raise_pep_call_exception at 0x109aa6c10>
__beartypistry = {'+6730761713772002360': (<class 'bool'>, <class 'NotImplementedType'>)}
args       = (<beartype_test.a00_unit.decor.code.test_codemypy.test_decor_mypy_notimplemented.<locals>.TheCloud object at 0x109ab1bb0>,
 <beartype_test.a00_unit.decor.code.test_codemypy.test_decor_mypy_notimplemented.<locals>.TheCloud object at 0x109ab1b80>)
kwargs     = {}

<string>:19: NameError
================================================================== short test summary info ===================================================================
FAILED beartype_test/a00_unit/decor/code/test_codemypy.py::test_decor_mypy_notimplemented - NameError: name 'NotImplementedType' is not defined
===================================================================== 1 failed in 0.21s ======================================================================
```

It's not obvious to me the source of this failure, or why the prior tests from 57e5e99 (still) succeed.

leycec:
**uhohes.** You just hit a `typing.ForwardRef` edge case with `@beartype` I *never* thought to test: forward references embedded in unions! That *shouldn't* be an issue, but anything untested is always an issue. So, this is an issue. `</sigh>`

Here's a minimal length example exhibiting the goodness:

```
# In "test_bear_cardio.py":
from beartype import beartype

NotImplementedType = type(NotImplemented)

class SeeBeartypeRun:
    def __init__(self, v: int) -> None:
        self.v = v

    @beartype
    def __add__(self, y: int) -> "C":
        if y == 1:
            return self
        return NotImplemented

c = C(4)
print(c + 1)
```

That's fine:

```
$ python3 test_bear_cardio.py
4
```

Here's a derivative example exhibiting the badness:

```
# In "test_bear_panic.py":
from beartype import beartype
from typing import Union

NotImplementedType = type(NotImplemented)

class SeeBeartypeRun:
    def __init__(self, v: int) -> None:
        self.v = v

    @beartype
    def __add__(self, y: int) -> Union["C", NotImplementedType]:
        if y == 1:
            return self
        return NotImplemented

c = C(4)
print(c + 1)
```

That's less fine:

```
$ python3 test_bear_panic.py
  File "test_case.py", line 34, in <module>
    print(c + 1)
  File "<string>", line 39, in __add__
NameError: name 'NotImplementedType' is not defined
```

And... that's awful. Whenever you hit this sort of snag, consider uncommenting line 149 of the private `beartype._util.func.utilfuncmake` submodule: e.g.,

```
        # print(f'\n@beartyped {func_name} wrapper:\n\n{number_lines(func_code)}\n')
```

Uncommenting that instructs `@beartype` to show the generated code for *all* type-checking wrappers, including the problematic one just above:

```
@beartyped __add__ wrapper:

(line 0001) def __add__(
(line 0002)     *args,
(line 0003)     __beartype_func=__beartype_func,
(line 0004)     __beartype_raise_exception=__beartype_raise_exception,
(line 0005)     __beartypistry=__beartypistry,
(line 0006)     **kwargs
(line 0007) ):
(line 0008)     # Localize the number of passed positional arguments for efficiency.
(line 0009)     __beartype_args_len = len(args)
(line 0010)     # Localize this positional or keyword parameter if passed *OR* to the
(line 0011)     # sentinel value "__beartype_raise_exception" guaranteed to never be passed.
(line 0012)     __beartype_pith_0 = (
(line 0013)         args[1] if __beartype_args_len > 1 else
(line 0014)         kwargs.get('y', __beartype_raise_exception)
(line 0015)     )
(line 0016) 
(line 0017)     # If this parameter was passed...
(line 0018)     if __beartype_pith_0 is not __beartype_raise_exception:
(line 0019)         # Type-check this passed parameter or return value against this
(line 0020)         # PEP-compliant type hint.
(line 0021)         if not isinstance(__beartype_pith_0, int):
(line 0022)             __beartype_raise_exception(
(line 0023)                 func=__beartype_func,
(line 0024)                 pith_name='y',
(line 0025)                 pith_value=__beartype_pith_0,
(line 0026)             )
(line 0027) 
(line 0028)     # Call this function with all passed parameters and localize the value
(line 0029)     # returned from this call.
(line 0030)     __beartype_pith_0 = __beartype_func(*args, **kwargs)
(line 0031) 
(line 0032)     # Noop required to artifically increase indentation level. Note that
(line 0033)     # CPython implicitly optimizes this conditional away - which is nice.
(line 0034)     if True:
(line 0035)         # Type-check this passed parameter or return value against this
(line 0036)         # PEP-compliant type hint.
(line 0037)         if not (
(line 0038)             # True only if this pith is of one of these types.
(line 0039)             isinstance(__beartype_pith_0, NotImplementedType) or
(line 0040)             isinstance(__beartype_pith_0, __beartypistry['__main__.C'])
(line 0041)         ):
(line 0042)             __beartype_raise_exception(
(line 0043)                 func=__beartype_func,
(line 0044)                 pith_name='return',
(line 0045)                 pith_value=__beartype_pith_0,
(line 0046)             )
(line 0047) 
(line 0048)     return __beartype_pith_0
```

This is the problematic line:

```
(line 0039)             isinstance(__beartype_pith_0, NotImplementedType) or
```

That shouldn't be happening. Beartype is smart enough to know when a type needs to be explicitly imported into the namespace of a type-checking wrapper. Sadly, it's not smart enough. Why? Because we handle builtins uniquely in `beartype`. Consider the `list` and `bool` builtin types. You don't have to explicitly import those; they're builtin, right? That's what "builtin" means. Right.

So, `beartype` automatically decides whether or not a type needs to be explicitly imported into the namespace of a type-checking wrapper based on whether or not that type resides in the `builtins` module. If it does, we don't import it, because we don't need to; else, we do.

So, what's the problem then? **The problem is that CPython frequently lies.** Check this madness out, for example:

```
$ ipython3
>>> NoneClass = type(None)
>>> f'{NoneClass.__module__}.{NoneClass.__name__}'
'builtins.NoneType'
>>> NoneType
NameError: name 'NoneType' is not defined
```

See what CPython just did there? It claims that `NoneType` (i.e., the class of the builtin `None` singleton) is a builtin. **It isn't.** That's a lie. There is *no* `builtins.NoneType`, even though CPython says there is.

The same exact thing is happening here. I foolishly assumed that `builtins.NoneType` was the only type that CPython lied about. Oh, pity me, for am I fool! Clearly, CPython is in fact lying about *many* types – including both `builtins.NoneType` and `builtins.NotImplementedType`, which is why your otherwise extremely well-articulated PR is failing.

The solution is for me to very quietly and even more quickly:

1. Find out just how many types CPython is lying about.
2. Refactor the (*waitforit*) private `beartype._util.clls.utilclstest.is_type_builtin()` tester to additionally handle those types. Currently, that tester only explicitly handles `NoneType`:

```
    if cls.__name__ == 'NoneType':
        return False
```

This is why the PEP 8 maxim of "Explicit is better than implicit" should have been followed more closely in the CPython interpreter itself. Since it wasn't, we now have to hack around the ill-advised missteps of Guido's youth. Curse thee with an unwashed bathtub full of quivering pseudopods, young Guido! :octopus:

posita:
Wow. That was more patience and guidance than any humble PR submitter could ever hope for. Grade A-1 professional collaboration with a generous dollop of friendly and humorous mentorship! And I've not yet sent you a fruitcake. That's how it's done, kids. Don't be fooled by all this arguing on the internet stuff going on. This is the Real Deal™, right here.

Let me give it a go and see how ~much damage I can do~ _far I can get_. At the outset, I took a stab at sussing out any other ~devious little liars~ _implementation details_. So far I came up with this:

```python
In […]: import builtins
    ...: from collections import defaultdict
    ...: types = defaultdict(list)
    ...: for name in dir(builtins):
    ...:   s = getattr(builtins, name)
    ...:   t = type(s)
    ...:   if t.__module__ == "builtins" and not hasattr(builtins, t.__name__):
    ...:     types[t].append((name, s))
    ...: for t in types:
    ...:   print(f"* {t.__module__}.{t.__name__}:\n  - ", end="")
    ...:   print(", ".join(name for name, _ in types[t]))
* builtins.ellipsis:
  - Ellipsis
* builtins.NoneType:
  - None
* builtins.NotImplementedType:
  - NotImplemented
* builtins.builtin_function_or_method:
  - __build_class__, __import__, abs, all, any, ascii, bin, breakpoint, callable, chr, compile, delattr, dir, divmod, eval, exec, format, getattr, globals, hasattr, hash, hex, id, input, isinstance, issubclass, iter, len, locals, max, min, next, oct, open, ord, pow, print, repr, round, setattr, sorted, sum, vars
* builtins.function:
  - display
* builtins.method:
  - get_ipython
```

The above does _**not**_ include things that make their way into `builtins` from other places:

```python
* _frozen_importlib.ModuleSpec:
  - __spec__
* _sitebuiltins._Printer:
  - copyright, credits, license
* _sitebuiltins._Helper:
  - help
```

Those all look legit, though.

As you probably guessed (or already know) `get_ipython` is an artifact of using [IPython](https://ipython.org/). However, its casual little inclusion serendipitously put `builtins.method` on the radar, which was kind. I tinkered around with types of various `class` entities, but was unable to surface anything other than `builtins.property` (legit), `builtins.builtin_function_or_method`, `builtins.function`, and `builtins.method`, which are already included above.

For funsies, I even tried the above technique in PyPy. I had _**hoped**_ such trickery was confined to CPython. Alas, my hopes were dashed, my soul imperceptibly darker from the impurities of leaky abstractions when I discovered it, too, contained _discrepancies_, not unlike CPython's:

```python
$ pypy37 --version
Python 3.7.9 (7e6e2bb30ac5fbdbd443619cae28c51d5c162a02, May 13 2021, 10:29:03)
[PyPy 7.3.3-beta0 with GCC Apple LLVM 12.0.0 (clang-1200.0.32.29)]
$ pypy37 <<EOF
import builtins
from collections import defaultdict
types = defaultdict(list)
for name in dir(builtins):
    s = getattr(builtins, name)
    t = type(s)
    if t.__module__ == "builtins" and not hasattr(builtins, t.__name__):
        types[t].append((name, s))
for t in types:
    print(f"* {t.__module__}.{t.__name__}:\n  - ", end="")
    print(", ".join(name for name, _ in types[t]))
EOF
* builtins.ellipsis:
  - Ellipsis
* builtins.NotImplementedType:
  - NotImplemented
* builtins.builtin_function:
  - __build_class__, abs, all, any, ascii, bin, breakpoint, callable, chr, compile, delattr, dir, divmod, eval, exec, format, getattr, globals, hasattr, hash, hex, id, input, isinstance, issubclass, iter, len, locals, max, min, next, oct, open, ord, pow, print, repr, round, setattr, sorted, sum, vars
* builtins.module:
  - __builtins__
* builtins.function:
  - __import__
```

At least with PyPy, there is no `builtins.method`. Methods are of type `builtins.function` (at least according to my casual inspection).

leycec:
Allow me to now emote: :exploding_head:

What an erudite analysis! I am now pinching myself. Surprisingly, I appear to be awake. This can only mean that you are indeed the reincarnated prophet of Computer Science born again as was long foretold on 4chan. I was just gonna manually iterate over the [published list of builtin constants](https://docs.python.org/3/library/constants.html), which would have been both insufficient *and* inefficient *and* shameful.

Instead, I now get to genuflect before your blatantly superior intellect. [My beautiful wife](https://www.researchgate.net/profile/Alexis-Pietak) will be sorely disappointed to learn of my failings, which is why we'll just keep this our little secret.

> And I've not yet sent you a fruitcake.

We expect our fruitcake to be much like the "top-soil" outside our ramshackle cottage on the [Canadian Shield](https://en.wikipedia.org/wiki/Canadian_Shield): rock-hard, full of colourful grit, and riddled with [questionable extremophiles](https://en.wikipedia.org/wiki/Extremophile).

> I took a stab at sussing out any other ~~devious little liars~~ *implementation details*.

That was such hot stuff that I stuffed everything you just pasted into a new testing-specific frozen set of all fake builtin types dynamically introspected from the `builtins` module:

```
TYPES_BUILTIN_FAKE = frozenset((
    # Type of this builtin.
    builtin.__class__
    # For each builtin (i.e., globally accessible object implicitly imported by
    # the active Python interpreter into *EVERY* lexical context)...
    for builtin in builtins.__dict__.values()
    # If...
    if (
        # The type of this builtin also insists itself to be defined by the
        # "builtins" module and thus also be a builtin *AND*...
        builtin.__class__.__module__ == MODULE_NAME_BUILTINS and
        # The "builtins" module contains *NO* globally-scoped attribute whose
        # name is that of this type...
        builtin.__class__.__name__ not in builtins.__dict__
    # Then add this cheatin', lyin', no-good type to this set.
    )
))
```

...*it works.* I know. I had to pinch myself again, which admittedly hurt more than expected. *Curse this papery thin skin.*

Given that as a sanity check, I then constructed the runtime-specific frozen set of all fake builtin types efficiently imported from the `beartype.cave`:

```
TYPES_BUILTIN_FAKE = frozenset((
    EllipsisType,
    FunctionType,
    FunctionOrMethodCType,
    MethodBoundInstanceOrClassType,
    NoneType,
    NotImplementedType,
))
```

...*it works, too.* We prefer the latter at runtime to minimize the cost of the first `import beartype` statement. We prefer the former at test-time as a sanity check against the latter, which will now automagically inform us when a new CPython release declares new fake builtins.

By the power of automagic, I compel thee to work, `@beartype`!

> At least with PyPy, there is no `builtins.method`. Methods are of type `builtins.function` (at least according to my casual inspection).

**Right!** PyPy is endlessly fascinating. You may already suspect why this is, but I shall now nonetheless pontificate boringly.

This is *probably* due to PyPy dynamically compiling all pure-Python callables (including bound methods) into machine code. Interestingly, this means that it's actually infeasible to differentiate callable types under PyPy. All callables have the same bloody type under PyPy, regardless of whether those callables were originally pure-Python or C-based.

Conversely, it's trivial to differentiate C-based functions from pure-Python functions from pure-Python bound methods under CPython just by inspecting their types, which CPython thankfully preserves. If you're curious and have a spare lifetime to burn, check out the oddly well-documented `"TYPES ~ call : function"` subsection of the poorly hidden `beartype._cave._cavefast` submodule.

I consider this a PyPy bug. PyPy developers consider me dumb. One of us is not wrong.

> The above does *not* include things that make their way into `builtins` from other places:

**Right!** We can only hope and pray to a burning effigy of Guido that there's a one-to-one correspondence between fake builtin types and builtin objects. As you've ingeniously discovered, all fake builtin types *appear* to originate from real builtin objects.

Let's assume for collective sanity that this is the case.

Lastly, **behold f293fa1!** The `is_type_builtin()` tester has now been refactored to avoid returning false positives for fake builtin types, which means that this PR *should* now pass. <sup>crossing tentacles</sup>

Would you mind rebasing this branch against `main` and repushing this PR? In deference to your l33t sk1llz, I tried harder than I usually try to preserve your hard work against merge conflicts. Pre-celebratory Canadian waffles for all! :waffle: :waffle: :waffle:

posita:
> I tried harder than I usually try to preserve your hard work against merge conflicts.

I fear not the conflicts of merge, so do as you will in times yet to come. However, I am grateful to be spared the battlefield this day.

> This can only mean that you are indeed the reincarnated prophet of Computer Science born again as was foretold long ago on 4chan.

Such prophesies carry weight too great for my shoulders to…_bear_. I am but a simpleton basking in the glory of this tentacley magic.

[“Two Fine Gentlemen” by Rebecca Flaum](https://www.etsy.com/listing/387092818/two-fine-gentlemen-8x10-art-print-with)

As such, I take no issue in swearing the oath of silence you require. ’Tis done. Also, you will be delighted to learn that my fruit cake ingredient shopping list now includes a casual trip to my nearest deep sea hydrothermal vent. This may increase time-to-delivery. Apologies for any inconvenience. I’m sure you understand.

I have rebased as you requested, which resulted in a new error that only required…how to put this…one, teensy tiny, surely insignificant, and not at all of the anti-pattern variety, mere [two word change](https://github.com/beartype/beartype/pull/40/files#diff-e69509666be41d5a5976d193fa9ace61fd1ce6e5ab52a8db6b59c643112fcac9R34-R38).

The _real_ Tentacley Prophet of the Science of Computation. Be wary of imposters. Accept no substitutes!

’Tis nothing really. Not even worth the bother of clicking the link above to see, especially in light of the glorious news that tests now pass. And what more evidence could—or _should_—be required than that?

leycec:
**ohnoes x 2.** The ~~tentacular horror~~ poor old bear never thought to test forward references to classes declared at callable rather than module scope. He hangs his muzzle in sadness.

I actually just blew through several hours and half a dozen steaming mugs of green tea – like Scarface-era Al Pacino on a bad FloridaMan night out – roughly forcing the poor old bear to support forward references to nested classes before reluctantly admitting that doing so would:

* Be crazy.
* Violate cultural norms (both domestic and foreign).
* Violate [PEP 563](https://www.python.org/dev/peps/pep-0563/) and Python ≥ 3.10, which effectively prohibit forward references to anything but globally scoped classes.
* Require non-portable and extremely inefficient inspection of the call stack by `@beartype`-generated wrapper functions at call time.

*Win some. Lose some.*

On the bright side, your PR is phenomenal, inscrutably funny, and overly poetic. <sup>luv it</sup> On the dark side, I now have less sanity and so possibly do you.

Let's merge this stunning brilliance before either of loses our tenuous grip on Internet-based reality.

leycec:
**Boom-shaka!** Thanks *so* much for the voluminous contribution, Matt. I profusely apologize for the two micro-issues you hit while resolving this for everyone.

Please don't hold that against the bear. He means well despite his advanced age, arthritic paws, and bad case of salmonitis.

posita:
> **Boom-shaka!** Thanks _so_ much for the voluminous contribution. I profusely apologize for the two micro-issues you hit while resolving this for everyone.
> 
> Please don't hold that against the bear. He means well despite his advanced age, arthritic paws, and bad case of salmonitis.

Apologies (profuse or otherwise) would imply that one or both of us stood in judgment of this collaboration. I—of course—honor any feelings of sadness you may have (just as I would any feelings, including, but not limited to those emanating from salmonitis), but know that my experience herein is filled entirely with joy and gratitude. I hope the exploration was fulfilling for you as well (and not just from the tea).

> Lastly, I hope to push out a new stable `beartype 0.7.2` release this Friday. I'll ping you when I do. Thanks for all the salmon, super-positive @posita.
> 
> P.S.: [dyce](https://github.com/posita/dyce) looks _super hot._ I'm hip-deep into board game theorycrafting, so a [native Python DSL for simulating dice rolls complete with overloaded matrix multiplication operator support](https://github.com/posita/dyce#a-taste) hits all the right cardboard chits. Yup! That's about to blow up. 💥

I am already working on the S-1 (Canadian translation: [Going public in Canada](https://duckduckgo.com/?q=%22going+public+in+canada%22), apparently).

My effort is but a mere toy in the shadow of the mighty bear, but it's fun (and hopefully useful to other hobbyists). Please don't hesitate to signal any deficiencies you might find. I would be thrilled for any excuse to work with you again.

PR #41, opened 2021-07-11
Re-comment debug print

posita:
Inadvertently left over from f293fa1d.

### DIFF
--- a/beartype/_util/func/utilfuncmake.py
+++ b/beartype/_util/func/utilfuncmake.py
@@ -146,7 +146,7 @@ def make_func(
     # Since doing so is both more verbose and obfuscatory for no tangible gain,
     # the current circumspect approach is preferred.
     try:
-        print(f'\n@beartyped {func_name} wrapper:\n\n{number_lines(func_code)}\n')
+        # print(f'\n@beartyped {func_name} wrapper:\n\n{number_lines(func_code)}\n')
         exec(func_code, func_globals, func_locals)
 
         #FIXME: See above.


leycec:
**Brilliant!** Okay... stunningly obvious, but I can only face-palm at having neglected to commit that myself. This is why they assuredly pay you the Big Bucks.™ Type-checking bears the world over remain grateful for the 1% of your juggernaut consciousness you've devote to their cause.

high-five bear proudly supports this PR
<sup>high-five proudly bear supports this PR</sup>

You might also be wondering when the next stable release of `beartype` (containing these generous improvements) will be dropping. The answer to that pertinent question and more is... as soon as I nail #42, a completely unrelated but incredibly critical feature request! Optimistic Cat predicts this should happen by the end of the week. :smile_cat:

posita:
> You might also be wondering when the next stable release of `beartype` (containing these generous improvements) will be dropping. The answer to that pertinent question and more is... as soon as I nail #42, a completely unrelated but incredibly critical feature request! Optimistic Cat predicts this should happen by the end of the week. 😸

I hope you don't feel any pressure to release on my behalf, but I appreciate the heads up.

### `<rant>`

By the way, I found the following passage (from [your Stack Exchange post](https://stackoverflow.com/questions/62871524/is-pep-585-unusable-at-runtime-under-python-3-7-and-3-8)) apt beyond type checking:

> In 2020, there are no good options – only a spectrum of increasingly horrifying lessers of several malignant evils. One would hope that PEP authors would actually test their implementations at runtime. Yet, here we are, adrift without a paddle in a steaming cesspit of theorycrafted anti-APIs. _Welcome to Python._

Some brief background (some of which you may have already surmised): Much of my exploration of the edges of [`beartype`](/beartype/beartype) emanates from my frustration with [PEP 3141](https://www.python.org/dev/peps/pep-3141/) and its ancestors. I wouldn't consider my effort significant (it's a hobby), but it does touch some similar pain points common to in-the-field data science and computation. Given the posturing in the standard library and various PEPs, it is remarkably and deceptively hard to build a computation library that can perform math on others' well-defined numerical primitives without breaking other features (like typing). My sense is that with solid APIs, it should be relatively straightforward. But it is not, which is especially troubling in light of the claims. And I am far from the first to [suffer cuts](https://github.com/python/mypy/issues/3186) from various sharp but latent edges. (Conversations documenting others' frustrations date back nearly a decade.)

In theory, as with typing, the evolution of Python's numerical primitives _appear_ as commitments (often with an air of confidence and excitement) around a common and consistent approach. In practice, it feels more like a disjointed set of inconsistencies that make sense only to wizards in towers disagreeing about implementation details rather than perfecting an ergonomic product for general use. The advertising invites exploration. The reality is that exploration—repeated in countless pockets of the world just like mine—is often in vain. Those who emerge from that journey have been taxed not only with their time, but their trust and enthusiasm. Perhaps that is our shared penance for having trusted the marketing, but that is deeply saddening.

Like you, S. Kirpichev [raises important questions](https://bugs.python.org/issue43602#msg391657) in his failed attempt to fix a particular wart. I note that (to the best of my knowledge) like yours, his questions remain tragically unanswered. Maybe hyper-specialized features like "numbers" and "typing" just don't have broad enough audiences to deserve the kind of attention and polish we hope for?

### `</rant>`

In any event, I join in [**Antyos**'s postscript](https://github.com/beartype/beartype/issues/42#issuecomment-880175750) and not only appreciate your prose, but applaud your persistence and maneuverability through hazard-strewn territory.

clapping bear

PR #51, opened 2021-09-29
Ensure README.rst is decoded as UTF-8 in test_doc_readme on all platforms

pbourke:
The default platform encoding on Windows is the single-byte encoding cp1252.
This causes an error when reading README.rst in test_doc_readme:

`"UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 1495: character maps to <undefined>"`

That is the 3rd byte in the unicode LEFT DOUBLE QUOTATION MARK character from the first code sample in the readme.

### DIFF
--- a/beartype_test/a90_func/doc/test_docreadme.py
+++ b/beartype_test/a90_func/doc/test_docreadme.py
@@ -48,7 +48,7 @@ def test_doc_readme(monkeypatch) -> None:
     from beartype_test.util.path.pytpathproject import get_project_readme_file
 
     # Decoded plaintext contents of this project's readme file as a string.
-    README_CONTENTS = get_project_readme_file().read_text()
+    README_CONTENTS = get_project_readme_file().read_text(encoding='utf-8')
 
     # List of all warning and error messages emitted by "docutils" during
     # parsing of this project's top-level "README.rst" file.


leycec:
`</facepalm>`

Thanks so much for both the investigative research and gruelling resolution, Pat. Your dedication to quality is applauded both here and at #52.

Everything passes, of course. Everything looks great, of course. Let's merge this! :+1:  

> cp1252

<sup>now that is a name i haven't heard in a long time</sup>

> That is the 3rd byte in the unicode LEFT DOUBLE QUOTATION MARK character from the first code sample in the readme.

I... I can't believe you know that. This must be why only one of us works for Microsoft.

Incidentally, I'm kinda surprised we got hit by this. Our GitHub Actions CI configuration assiduously tests against Windows...

**OH. I SEE.** [We intentionally run `tox` under `bash` rather than a native Windows shell](https://github.com/beartype/beartype/blob/2b1f3653d97eb29d1b59e2d409d863004a30e39d/.github/workflows/python_test.yml#L189), because we have to conditionally kludge package installation under MacOS to account for their broken Accelerate library. Apple always gonna Apple.

pbourke:
> I... I can't believe you know that.

Of course - I lull myself to sleep every night with the Basic Multilingual Plane, as one does

Incidentally you can do `:%!xxd` in vim to get a hex dump of your buffer. It even works in Windows - the vim distribution includes `xxd`.

leycec:
> Of course - I lull myself to sleep every night with the Basic Multilingual Plane, as one does

*flashbacks of reading the unicode standard for nebulous fun intensify*

> Incidentally you can do `:%!xxd` in vim to get a hex dump of your buffer.

Shamefully... I did not know that. Thankfully, `xxd` is only two character transforms away from [xkcd](https://xkcd.com/)! I may possibly remember this by mnemonic. When I fail to, I'll grep GitHub until exhaustedly giving up and regretting not bookmarking this wisdom.

> It even works in Windows...

But... but doesn't *everything* work in Windows? I only jest, of course. We love Microsoft here! <sup>please don't delete my repo, please don't delete my repo</sup>

PR #70, opened 2021-11-14
Update boto3 docs to be more accurate

paulhutchings:
I feel ashamed to say that my original implementation of beartype validators for boto3 types was... _(gasp, shock)_ 

_incorrect_

TLDR, it appears that any kind of custom annotated type that uses just `object` as its base will not be properly validated. Thankfully I was able to find base classes in the botos that when used will ensure correct checking of custom types. And [tests](https://github.com/paulhutchings/bearboto3/blob/s3/tests/s3/test_s3.py) to prove it too! (Thank goodness for TDD or I may never have discovered this problem).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -614,6 +614,7 @@ Validators_>`__ to satisfy both static and runtime type checkers:
    # Import the requisite machinery.
    from beartype import beartype
    from boto3 import resource
+   from boto3.resources.base import ServiceResource
    from typing import TYPE_CHECKING
 
    # If performing static type-checking (e.g., mypy, pyright), import boto3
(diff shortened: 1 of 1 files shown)


leycec:
> I feel ashamed to say that my original implementation of beartype validators for boto3 types was... *(gasp, shock)*

oh.my.god

Also, [`bearboto3`](https://github.com/paulhutchings/bearboto3) is a *thing*! Praise be to you, Paul, and the extend family tree that eventually begat you. Heroes one and all.

Also, a double-shocker emerges from the `bearboto3` docos. I *think* a river dolphin with the head of a bear <sup>...it's *beautiful*</sup> may have accidentally eaten a word between the "will" and the "since":

> However, if you are using a runtime type-checker such as `beartype`, type validation will **[something, something, fail, something, something]** since the types technically do not match (even though they represent the same object schema).

Lastly, drop me a ping when you've released your first stable version on PyPI and/or `conda-forge`. I'll shill, stan, and namedrop hard for you and `bearboto3` – or my real-world nickname isn't Lazy Boy.

> Thankfully I was able to find base classes in the botos that when used will ensure correct checking of custom types.

I... I didn't even know the botos *had* classes. Like, any classes. This is truly a day of mass celebration. "Raise those keyboarding hands in the air, like Amazon just don't care!" :raised_hands:

leycec:
Oh, *wonderful.* You already helpfully splashed `bearboto3` across our docos! That's great. Quite a weight off my geriatric elderly millennial mind, really. **It's in.**

I'll lightly edit this to give `bearboto3` top billing. Anybody that got time for both Boto3 and `beartype` also got time for `bearboto3`. Since this is non-trivial and fragile, everybody should go straight to the guy who knows what he's doing for once. Thanks again for all the QA goodies, Paul! Also, please make a river dolphin with the head of a bear happen.

leycec:
...let's pretend I just didn't actually close this without merging. `</awkward_moments>`

leycec:
...let's pretend I just didn't actually close this without merging. `</awkward_moments>`

PR #86, opened 2022-01-22
Initial attempt a caching protocol implementation

posita:
Initial stab at porting [``numerary.types.CachingProtocolMeta``](https://github.com/posita/numerary/blob/dd57bffff32c8e1c4d46409c9302318bdcfe3e20/numerary/types.py#L129) to `beartype`. It only implements a naive cache that will grow indefinitely. (I.e., it does not include an LRU mechanism.) That can be addressed in a subsequent PR (before this one lands, if necessary). Further, this port does _not_ include `numerary`'s `include`/`exclude` mechanism, which _should_ keep this implementation simpler.

Eventually, this _should_ fix posita/numerary#8.


beartype/typing.py
```diff
@@ -288,3 +280,197 @@
     Set = set  # type: ignore[misc]
     Tuple = tuple  # type: ignore[assignment]
     Type = type  # type: ignore[assignment]
+
+if _IS_PYTHON_AT_LEAST_3_8:
+    # TODO: I'm thinking maybe this belongs in its own file that is imported
+    # here
```
posita, line 286: Maybe this should go into a peer module (e.g., `protocol.py`)?

beartype/typing.py
```diff
@@ -288,3 +280,197 @@
...
+    _TT = TypeVar("_TT", bound="type")
+
+    if TYPE_CHECKING:
+        # Warning: Deep typing voodoo ahead. See
+        # <https://github.com/python/mypy/issues/11614>.
+        from abc import ABCMeta as _ProtocolMeta
+    else:
+        _ProtocolMeta = type(_Protocol)
+
+    # TODO: Rename this?
```
posita, line 305: I kept the leading `_` for now to not run afoul of the name checking in `test_api_typing.py`. I'm not quite sure I understand what's going on there or how to adapt to it, but we can discuss….

beartype/typing.py
```diff
@@ -288,3 +280,197 @@
...
+
+if _IS_PYTHON_AT_LEAST_3_8:
+    # TODO: I'm thinking maybe this belongs in its own file that is imported
+    # here
+
+    # Define our own caching protocol
+    from typing import Protocol as _Protocol
+    from typing import runtime_checkable as runtime_checkable
+    # TODO: This is...a wart. Obviously, there is a fragility here. We should
+    # discuss.
```
posita, line 292: This may be a larger topic…. 😬

beartype/typing.py
```diff
@@ -288,3 +280,197 @@
...
+          >>> my_thing: MyBearProtocol = MyImplementation()
+          >>> isinstance(my_thing, MyBearProtocol)
+          True
+
+          >>> @beartype
+          ... def do_somthing(thing: MyBearProtocol) -> None:
+          ...   thing.myfunc(0)
+        """
+        __slots__: Union[str, Iterable[str]] = ()
+
```
posita, line 439: From here on, I create caching versions of the standard library's `Supports` protocols. This might be taking liberties, in which case I can remove them. If they should stay, I'll probably need some guidance on how to overcome the current type name checking test failures.

beartype/typing.py
```diff
@@ -288,3 +280,197 @@
...
+            # See <https://github.com/python/mypy/issues/9282>
+            cls = super().__new__(mcls, name, bases, namespace, **kw)  # type: ignore [misc]
+            # Prefixing this class member with "_abc_" is necessary to prevent
+            # it from being considered part of the Protocol. (See
+            # <https://github.com/python/cpython/blob/main/Lib/typing.py>.)
+            cls._abc_inst_check_cache = {}
+
+            return cls
+
+        def __instancecheck__(cls, inst: Any) -> bool:
```
posita, line 385: This implementation is ported straight across from `numerary` and does not limit entries, which we'll likely want to introduce as part of this PR. I thought treating this as a functioning placeholder was a good place to start, though.

posita (commented):
Highlighted some questions/areas of concern.

posita:
Oh! Let me know if you prefer that I keep my commit(s) tidy (i.e., squash/rewrite as a go), or whether you prefer to preserve a longer commit chain here and tidy up at the end. (Different maintainers have different preferences, so I want to make sure I don't mess up your flow.)

posita (commented):
Houston, we have a hiccup.¹ I forgot (or willed from my mind) some thorny issues around `Protocol`s that may frustrate one of the goals of this PR, namely: Creating a drop-in replacement for `typing.Protocol`.

I'm _not_ an expert, but as far as I can tell, `typing.Protocol` is a special animal. _Really_ special. Like Chimera special.

For Mypy to work, I'm pretty sure it needs to [think it's dealing with things _directly_ derived from `typing.Protocol` and nothing else](https://mypy.readthedocs.io/en/stable/protocols.html#defining-subprotocols-and-subclassing-protocols). _Buuut_ for a world where others can gain access to caching behavior through mere subclassing, they need a parent with our metaclass implementation, which almost certainly isn't `typing.Protocol`.

~~I'm pretty sure the [first import of `typing.Protocol` as `Protocol`](https://github.com/beartype/beartype/blob/82160299cbf92d63d716890e1fa65d30a57e8e59/beartype/typing.py#L161) "hides" our implementation from Mypy (hence the various `# type: ignore [no-redef]` comments), which tricks it into allowing static checking to proceed as normal. I think this is why we _don't_ get static type-check errors for `test_api_typing.py`, even though it uses non-compliant "protocols" (i.e., those that derive from `beartype.typing.Protocol` without also deriving from `typing.Protocol`) all over the place.~~

~~That's a happy accident … _for now_. (I have no idea if that holds for other static type checkers.)~~

_UPDATE: I believe I have settled on a solution (see my [subsequent comment](https://github.com/beartype/beartype/pull/86#issuecomment-1020202249)) with one minor compromise by putting implementations in their own module (`_protocol.py` for now) and conditionally importing them into `typing.py`._

The _bigger_ issue is [generic protocols](https://mypy.readthedocs.io/en/stable/generics.html#generic-protocols). As far as I know, one can't have arbitrary generic placeholders. But `typing.Protocol` _can_. It behaves like `typing.Generic` in that regard. In other words, you can totally do this:

``` python
from abc import abstractmethod
from typing import Protocol, TypeVar, runtime_checkable
_AT_co = TypeVar("_AT_co", covariant=True)
_BT_co = TypeVar("_BT_co", covariant=True)
…
_NT_co = TypeVar("_NT_co", covariant=True)

@runtime_checkable
class SupportsOneTwo(Protocol[_AT_co, _BT_co, …, _NT_co]):  # <-- WTF?!
    @abstractmethod
    def a(self) -> _AT_co:
        pass
    @abstractmethod
    def b(self) -> _BT_co:
        pass
    …
    @abstractmethod
    def n(self) -> _NT_co:
        pass
```

What I don't know how to do is to replicate `typing.Protocol`'s ability to subscript arbitrary type parameters in our own class without requiring subclasses explicitly include `Generic`. So one could do this …

``` python
from typing import Generic
from beartype.typing import Protocol

class SupportsOneTwo(Protocol, Generic[_AT_co, _BT_co, …, _NT_co]):
    ...
```

… or this …

``` python
from typing import Protocol, runtime_checkable
from beartype.typing import _CachingProtocolMeta

@runtime_checkable
class SupportsOneTwo(Protocol[_AT_co, _BT_co, …, _NT_co], metaclass=_CachingProtocolMeta):
    ...
```

… or this …

``` python
import typing
from beartype.typing import Protocol

class SupportsOneTwo(Protocol, typing.Protocol[_AT_co, _BT_co, …, _NT_co]):
    ...
```

… but one cannot do this …

``` python
from beartype.typing import Protocol

class SupportsOneTwo(Protocol[_AT_co, _BT_co, …, _NT_co]):
    ...
```

You'll get a module loading error claiming that `beartype.typing.Protocol` is not generic. I'm stumped on how to get a syntax-compatible version. My hunch is that the best thing we can do in this case is tell customers to use `Generic` and not `Protocol` (i.e., the first alternative I presented above).

---
¹ Or "hiccough", if one prefers it. 🧐

beartype/typing.py
```diff
@@ -288,3 +288,211 @@
...
+            for attr in attrs:
+                if not hasattr(inst, attr):
+                    return False
+                elif callable(getattr(cls, attr, None)) and getattr(inst, attr) is None:
+                    return False
+
+            return True
+
+    @runtime_checkable
+    class Protocol(_Protocol, Generic, metaclass=_CachingProtocolMeta):  # type: ignore [no-redef]
```
posita, line 435: Just a heads up: I don't think deriving from `Generic` here does anything. It was an experiment to see if I could work through the issue I raised in [this comment](https://github.com/beartype/beartype/pull/86#pullrequestreview-860250362).

posita:
UPDATE: Asking a related question [here](https://gitter.im/python/typing?at=61eca4a0d143b14f832e6b2f).

beartype/typing.py
```diff
@@ -288,3 +288,211 @@
...
+    from typing import SupportsFloat as _SupportsFloat
+    from typing import SupportsIndex as _SupportsIndex
+    from typing import SupportsInt as _SupportsInt
+    from typing import SupportsRound as _SupportsRound
+
+    class SupportsAbs(_SupportsAbs[_T_co], Protocol, Generic[_T_co]):  # type: ignore [no-redef]
+        "A caching version of :class:`typing.SupportsAbs`."
+        __slots__: Union[str, Iterable[str]] = ()
+
+    class SupportsBytes(_SupportsAbs, Protocol):  # type: ignore [no-redef]
```
posita, line 476: Whoops! This should be `class SupportsBytes(_SupportsBytes, Protocol): …`. I'll fix this….

beartype/typing.py
```diff
@@ -288,3 +288,211 @@
...
+    _TT = TypeVar("_TT", bound="type")
+
+    if TYPE_CHECKING:
+        # Warning: Deep typing voodoo ahead. See
+        # <https://github.com/python/mypy/issues/11614>.
+        from abc import ABCMeta as _ProtocolMeta
+    else:
+        _ProtocolMeta = type(_Protocol)
+
+    # TODO: Rename this?
```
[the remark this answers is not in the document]
posita: This has been moved to `_protocol.py` in the latest comment, but the issue of what to call it remains.

beartype/typing.py
```diff
@@ -288,3 +288,211 @@
...
+          >>> my_thing: MyBearProtocol = MyImplementation()
+          >>> isinstance(my_thing, MyBearProtocol)
+          True
+
+          >>> @beartype
+          ... def do_somthing(thing: MyBearProtocol) -> None:
+          ...   thing.myfunc(0)
+        """
+        __slots__: Union[str, Iterable[str]] = ()
+
```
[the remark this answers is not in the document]
posita: These live in `_protocol.py` in the latest commit, but the issue of whether to include them at all remains.

beartype/typing.py
```diff
@@ -288,3 +288,211 @@
...
+            cls = super().__new__(mcls, name, bases, namespace, **kw)  # type: ignore [misc]
+            # Prefixing this class member with "_abc_" is necessary to prevent
+            # it from being considered part of the Protocol. (See
+            # <https://github.com/python/cpython/blob/main/Lib/typing.py>.)
+            cls._abc_inst_check_cache = {}
+
+            return cls
+
+        def __instancecheck__(cls, inst: Any) -> bool:
+            try:
```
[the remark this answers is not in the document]
posita: The implementation lives in `_protocol.py` as of the latest commit, but a caching implementation still needs to be considered.

beartype/typing.py
posita: Corrected as of a35e6de9edce0cdbfc76fe047bd9823ec0d1bfd8.

posita:
6ad8629f73aea00403589bb636772342ce8df9c6 pulls out the implementation into `_protocol.py`. I *think* this is cleaner. There are some redundant guarded imports, but that's part of the strategy. The idea is that `_protocol.py` conditionally creates the overrides, but in such a way that `_protocol.py` itself can be type-checked. `typing.py` handles the `TYPE_CHECKING`-guarded `import`s such that Mypy doesn't see beartype's overrides (for reasons alluded to in [my prior comment](https://github.com/beartype/beartype/pull/86#pullrequestreview-860250362)). This is important to so that it treats `beartype.typing.Protocol` identically to `Protocol`. (The same goes for the various `Supports` protocols.)

The *one* wart here is that Mypy won't detect this problem:

``` python
>>> from beartype.typing import Protocol, TypeVar
>>> _T = TypeVar("_T")

>>> class MyProtocol(Protocol[_T]):  # <-- can't do this
>>>   pass
```

Users of `beartype.typing.Protocol` have to do this instead:

``` python
>>> from beartype.typing import Generic, Protocol, TypeVar
>>> _T = TypeVar("_T")

>>> class MyProtocol(Protocol, Generic[_T]):
>>>   pass
```

beartype/typing.py
```diff
@@ -288,3 +288,211 @@
...
     Tuple = tuple  # type: ignore[assignment]
     Type = type  # type: ignore[assignment]
+
+if _IS_PYTHON_AT_LEAST_3_8:
+    # TODO: I'm thinking maybe this belongs in its own file that is imported
+    # here
+
+    # Define our own caching protocol
+    from typing import Protocol as _Protocol
+    # TODO: This is...a wart. Obviously, there is a fragility here. We should
```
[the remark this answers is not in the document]
posita: New location: https://github.com/beartype/beartype/pull/86/files#diff-7f21004698cf70eabd8a0424eb9e87d0501132ee48aeb9c7011c6356fdca52ddR35-R36

beartype/typing.py
```diff
@@ -288,3 +288,211 @@
     Set = set  # type: ignore[misc]
     Tuple = tuple  # type: ignore[assignment]
     Type = type  # type: ignore[assignment]
+
+if _IS_PYTHON_AT_LEAST_3_8:
+    # TODO: I'm thinking maybe this belongs in its own file that is imported
```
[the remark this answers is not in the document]
posita: Made this `_protocol.py` for now.

beartype/typing.py
posita: Removed.

beartype/_protocol.py
```diff
@@ -0,0 +1,236 @@
+# --------------------( LICENSE                           )--------------------
+# Copyright (c) 2014-2022 Beartype authors.
+# See "LICENSE" for further details.
```
posita, line 3: I know you have particular stylistic preferences for organization/naming/commenting. If you're okay with it, I won't attempt to meet all those preferences here. Rather, I'll focus on getting the rough structure in place first.

leycec:
This right here **straight-up Naruto run.**

i advise against trying this

Yet again, @posita went above and beyond the bear call of duty. Aren't bears supposed to sleeping in a fitful hibernate slumber filled with berry bunches, bloodied muzzles, and swaggering female bear haunches this time of year anyway!?!?

Yet again, I am [beyond the wall of sleep](https://www.hplovecraft.com/writings/texts/fiction/bws.aspx) myself. This means I will remain verbose yet disconcertingly incoherent. Let's begin.

> I know you have particular stylistic preferences for organization/naming/commenting.

*...eye spasmodically twitches.*

**Blackest heresy!** There is only *one* stylistic preference – and that is mine, I believe. All others are merely accidents their authors have yet to openly acknowledge. I'm kidding! Really. I'm not the unseemly jerk our neighbours continually claim me to be (however factual their video evidence may be).

Alternately, I'm obsessive-compulsive to the detriment of every codebase within keyboard reach of my pudgy fingers. I dispute neither claim.

> If you're okay with it, I won't attempt to meet all those preferences here.

*...very well.* We reserve the Comfy Chair for another occasion then.

yes, oh yes

> 6ad8629 pulls out the implementation into `_protocol.py`.

You have done well, "young" protégé. I see your code is as big as mine.

Seriously, though. This rocks. For those who are about to code, I salute you and then genuflect until my back cracks in every single vertebrae simultaneously.

> `TYPE_CHECKING`

Checkmate, mypy. Checkmate.

> This implementation is ported straight across from `numerary`...

vigorous *nod nod nod*

> ...and does not limit entries.

**...wut u say!?!?**

> I thought treating this as a functioning placeholder was a good place to start, though.

`</ahem>`

You are, of course, always correct.

vigorous *nod nod nod*

> I kept the leading `_` for now to not run afoul of the name checking in `test_api_typing.py`. I'm not quite sure I understand what's going on there or how to adapt to it, but we can discuss….

*...heh.* In hindsight, I probably should have documented that. With only minimal hand-waving, `test_api_typing.py` exists to exercise the constraint that there exists <sup>*...just get to it already*</sup> a one-to-one-mapping between public exported attributes of `beartype.typing` and `typing`. If `typing` exports something that `beartype.typing` does *not*, `test_api_typing.py` raises a test failure.

I think. Maybe. Actually, let's not quote me on that.

> You'll get a module loading error claiming that `beartype.typing.Protocol` is not generic.

**Hindu elephant deity Ganesha preserve us.**

That smells suspiciously like a poo-smelling deal-breaker. Let's see if we can't resolve this without any further reference to the Comfy Chair. Are you perhaps hitting this exception buried within the bowels of `typing.py`?

``` python
def _check_generic(cls, parameters, elen):
    """Check correct count for parameters of a generic cls (internal helper).
    This gives a nice error message in case of count mismatch.
    """
    if not elen:
        # v--- THIS BAD DUDE, RIGHT?
        raise TypeError(f"{cls} is not a generic class")
    alen = len(parameters)
    if alen != elen:
        raise TypeError(f"Too {'many' if alen > elen else 'few'} arguments for {cls};"
                        f" actual {alen}, expected {elen}")
```

Since `typing.Generic` and `typing.Protocol` are pure-Python, we should (*crossing pudgy fingers that kinda just rub against one another rather than crossing*) be able to adroitly mimic the hideous incantation they're casting. Shut your eyes, Marion!

With the well-honed intuition of one who started violating privacy encapsulation when he was only 5, I intuit that the core issue is that <sup>*...just get to it already*</sup> the `Generic.__class_getitem__()` dunder method *not* redefined by `Protocol` hardcodes itself and `Protocol` in an unsightly `if` conditional that makes that burst capillary in my eye throb with bloodshot viscosity even harder than it used to:

``` python
    @_tp_cache
    def __class_getitem__(cls, params):
        if not isinstance(params, tuple):
            params = (params,)
        if not params and cls is not Tuple:
            raise TypeError(
                f"Parameter list to {cls.__qualname__}[...] cannot be empty")
        params = tuple(_type_convert(p) for p in params)
        if cls in (Generic, Protocol):  # <-- THIS BAD DUDE RIGHT HERE
            # Generic and Protocol can only be subscripted with unique type variables.
            if not all(isinstance(p, (TypeVar, ParamSpec)) for p in params):
                raise TypeError(
                    f"Parameters to {cls.__name__}[...] must all be type variables "
                    f"or parameter specification variables.")
            if len(set(params)) != len(params):
                raise TypeError(
                    f"Parameters to {cls.__name__}[...] must all be unique")
        else:
            # Subscripting a regular Generic subclass.
            if any(isinstance(t, ParamSpec) for t in cls.__parameters__):
                params = _prepare_paramspec_params(cls, params)
            else:
                _check_generic(cls, params, len(cls.__parameters__))
        return _GenericAlias(cls, params,
                             _typevar_types=(TypeVar, ParamSpec),
                             _paramspec_tvars=True)
```

Let's circumvent that. Now hear me out here. Things gonna get ugly. Ever seen *Reservoir Dogs*? We're reproducing the ending right now.

``` python
class Protocol(...):
    ...

    def __class_getitem__(cls, params):
        super().__class_getitem__(_Protocol, params)  # <-- MAD CACKLING HERE
```

Pretty sure that'll work – but dead certain that'll fail. When it does, we consider options. Options that are labelled "DO NOT PUSH THIS SICKENINGLY RED BUTTON" include:

* Redefining the entirety of `Generic.__class_getitem__()` as our own `Protocol.__class_getitem__()` that just stupidly replaces that stupid `if cls in (Generic, Protocol):` conditional with our homebrew `if cls in (Generic, GodsHelpTypingAcceptOurSuperiorProtocolImplementationAlready):`.
* Temporarily monkey-patching `typing.Protocol` for the duration of that call. We are now literally reproducing *Reservoir Dogs* in Python by doing this, but we claim that `typing` leaves us no choice:

``` python
class Protocol(...):
    ...

    def __class_getitem__(cls, params):
        Prococol_old = typing.Protocol
        typing.Protocol = cls  # <-- MAD CACKLING HERE
        item = super().__class_getitem__(cls, params)  # <-- HERE AS WELL
        typing.Protocol = Protocol_old  # <-- CACKLING STILL AUDIBLE
        return item
```

this is so me

posita:
Thank you for the education on [`object.__class_getitem__`](https://docs.python.org/3/reference/datamodel.html#object.__class_getitem__)! Consider my mind sufficiently blown.

[Well there's something you don't see every day.](https://youtu.be/mj4_Km11F0Q)

I ain't too proud to beg [a system to do what I want by runtime patching¹]. At the risk of hitting the jolly, candy-like [history eraser button](https://vimeo.com/126720159), after looking at the [implementation of `Generic.__class_getitem__`](https://github.com/python/cpython/blob/072f4a473e861c6c987650f08990c0ed1f76715f/Lib/typing.py#L1301-L1326), I'm curious about the practical difference between something like this (adapted from above) …

``` python
class Protocol(...):
    ...

    def __class_getitem__(cls, params):
        Protocol_old = typing.Protocol
        try:
            typing.Protocol = cls
            return super().__class_getitem__(params)
        finally:
            typing.Protocol = Protocol_old
```

… and this (also adapted from above) …

``` python
class Protocol(...):
    ...

    def __class_getitem__(cls, params):
        _Protocol.__class_getitem__(params)
```

… ?²

One remote problem with the first approach _**might**_ be a lack of thread safety. I assume that doesn't come up very often, but who knows? Something something … web frameworks … something something … runtime (re)loading of packages in separate threads … something something … dependency injection … [remainder deleted in the interest of time].

I _**think**_ both encounter the [LRU cache](https://github.com/python/cpython/blob/072f4a473e861c6c987650f08990c0ed1f76715f/Lib/typing.py#L303-L323) while masquerading as `typing.Protocol`. I don't know that I fully understand the implications. I'm also not confident I fully understand that implementation, but it _**seems**_ like there aren't any side effects with `cls`, just what gets stuffed into the returned `_GenericAlias` (which will necessarily be `typing.Protocol` rather than `beartype.typing.Protocol`). I don't fully appreciate what that means either.³

That being said, I see your _Reservoir Dogs_ ending and raise you a _Taxi Driver_ with:

``` python
class Protocol(_Protocol, …):
    # …
    def __class_getitem__(cls, params):
        gen_alias = _Protocol.__class_getitem__(params)
        return type(gen_alias)(cls, *gen_alias.__args__)
```

---

¹ I concede that starts to sound less like _begging_ and more like _coercing_, but let's leave that be for the time being.

² Mad cackling omitted for brevity, but assume it persists throughout the quoted material.

³ I have a lot of skepticism around my own knowledge in this area, if you didn't catch that theme.

beartype/typing.py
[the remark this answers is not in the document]
posita: Resolving for now. We can deal with naming conventions later.

beartype/_protocol.py
posita: Putting the comfy chair back for now. Happy to ~revisit~ _continue to avoid_ later as necessary. 😜

beartype/_protocol.py
```diff
@@ -0,0 +1,245 @@
...
+        __slots__: Union[str, Iterable[str]] = ()
+
+        def __class_getitem__(cls, params):
+            gen_alias = _Protocol.__class_getitem__(params)
+
+            # I'm pretty sure we need this nudge. Otherwise our inheritors end
+            # up with the wrong metaclass (i.e., type(typing.Protocol) instead
+            # of the desired _CachingProtocolMeta). Luddite alert: I don't
+            # fully understand the mechanics here.
+            return type(gen_alias)(cls, *gen_alias.__args__)
```
posita, line 217: Whoo boy.

beartype/typing.py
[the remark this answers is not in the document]
posita: If we're going through the trouble of creating a drop-in caching replacement for `Protocol`, I'm going to assume we'll be successful and also assume that we would want caching versions of the standard library `Supports` protocols to go with it (and maybe even others).

It would seem silly to me to require a client who went through the trouble of typing `from beartype.typing import Protocol, SupportsIndex` to do _extra_ work to ensure the `SupportsIndex` they just imported got the benefits of the `Protocol` they just imported.

Resolving for now, but we can revisit if needed.

beartype/_protocol.py
```diff
@@ -0,0 +1,245 @@
...
+         TYPE_CHECKING,
+         Any,
+         Generic,
+         TypeVar,
+         Union,
+         runtime_checkable,
+    )
+
+    # Stuff we're going to override herein
+    from typing import (  # type: ignore[attr-defined]
```
posita, line 24: I'm not sure why Mypy chokes on this in CI for Python 3.7.

beartype/typing.py
```diff
@@ -288,3 +281,43 @@
...
+        SupportsAbs as SupportsAbs,
+        SupportsBytes as SupportsBytes,
+        SupportsComplex as SupportsComplex,
+        SupportsFloat as SupportsFloat,
+        SupportsInt as SupportsInt,
+        SupportsRound as SupportsRound,
+    )
+
+    if _IS_PYTHON_AT_LEAST_3_8:
+        from typing import (  # type: ignore[attr-defined]
```
posita, line 297: I'm not sure why Mypy chokes on this in CI for Python 3.7.

leycec:
**This is amazing fiery balls,** if you hadn't already noticed. :sun_with_face: 

But first!

> I'm not sure why Mypy chokes on this in CI for Python 3.7.

Because...

this is mypy

Because mypy is just a pseudonym for He Who Walks Behind the Rows, let's pretend mypy isn't watching like a voyeur in the corn fields for now. As time permits and the cats finally stop howling for food, we can uppercut mypy's glass jaw with a spray of `# type: ignore[EVERYTHING. IGNORE EVERYTHING, MYPY.]` comments in `beartype.typing`.

For now, we glare sternly in mypy's general direction.

But next!

``` python
       def __class_getitem__(cls, params):
            gen_alias = _Protocol.__class_getitem__(params)

            # I'm pretty sure we need this nudge. Otherwise our inheritors end
            # up with the wrong metaclass (i.e., type(typing.Protocol) instead
            # of the desired _CachingProtocolMeta). Luddite alert: I don't
            # fully understand the mechanics here.
            return type(gen_alias)(cls, *gen_alias.__args__)
```

You have now gone to a place where none but eagles dare.

You *almost* got there without burning your microlight wings made of greasy (yet patented) paraffin wax in the positronic radiation field emitted by the standard `typing` module – sorta like how I *almost* made my wife proud by cross-country skiing only five times as slow as her today. It's a special milestone for me.

The burning concern here is the instantiation of `_GenericAlias`. We're dropping a few parameters we really shouldn't be. Now the shambling horde of unkillable zombies whose only weakness is a blowtorch are gathering about the *absolutely* safe first-floor mall food court we're desperately pilfering for penicillin-encrusted snow cones and last month's sushi.

Under Python 3.10, here's how `Protocol` crazily instantiates `_GenericAlias`:

``` python
    @_tp_cache
    def __class_getitem__(cls, params):
        ...
        return _GenericAlias(cls, params,
                             _typevar_types=(TypeVar, ParamSpec),
                             _paramspec_tvars=True)
```

So `Protocol` is instantiating `_GenericAlias` with non-default values for both `_typevar_types` and `_paramspec_tvars`. I *definitely* know *exactly* what those parameters signify but have omitted an extensive explanation due to... space constraints. Yes. Space constraints, I'm afraid. <sup>*nervous hand wringing*</sup>

I'm now squinting suspiciously at `*gen_alias.__args__`, too. If `_GenericAlias.__init__()` can be trusted, <sup>*nervous eye twitching*</sup> `*gen_alias.__args__` should *basically* be equivalent to `*params` – which definitely doesn't seem right, because:

* `params` should be passed non-variadically. I think. Maybe. Actually, I have no idea what's happening here anymore. When nothing has meaning, everything has meaning.
* We're not passing the right non-default values for either `_typevar_types` or `_paramspec_tvars`. **Caveat emptor:** I can barely remember my own name at this point. I'm pretty sure it's Yojimbo Baggins... *but not certain.*

Because I'm now squinting suspiciously at everything, let's slowly backtrack to the last point in the autumn woods *not* inhabited by unsettling effigies, cawing crows, and a furtive shadow at the corner of your eye that keeps shifting even as you uneasily pretend "...it's really not there, it's not really there."

This is that point:

``` python
class Protocol(...):
    ...

    def __class_getitem__(cls, params):
        Protocol_old = typing.Protocol
        try:
            typing.Protocol = cls
            return super().__class_getitem__(params)
        finally:
            typing.Protocol = Protocol_old
```

this is galaxy brain

**This is galaxy brain.** Unlike my plebeian proposal, your upper-class approach cleverly accounts for unexpected exceptions.

Let's rock this like Rosemary's baby in the cradle.

beartype/_protocol.py
```diff
@@ -0,0 +1,277 @@
...
+                new_gen_alias._typevar_types = fake_gen_alias._typevar_types
+            if hasattr(fake_gen_alias, "_paramspec_tvars"):
+                new_gen_alias._paramspec_tvars = fake_gen_alias._paramspec_tvars
+
+            # Now all we have to do is replace the __origin__ in our clone to
+            # point to ourself.
+            new_gen_alias.__origin__ = cls
+
+            # We're done! Time for a honey brewskie break. We earned it.
+            return new_gen_alias
```
posita, line 249: One last attempt….

posita:
> This is that point: … Let's rock this like Rosemary's baby in the cradle.

Your wish is my command…_eventually_. Allow me first to draw attention to one (perceived) flaw of Rosemary's Galaxy Baby, which may amount to a [nothingburger](https://www.merriam-webster.com/words-at-play/nothing-burger-nothingburger-meaning-origin-phrase), as well as one last delve into my misguided hijack-the-generic-alias approach.

## Theoretical race condition present in the we're-going-to-replace-that-faberge-egg-with-a-fake-but-just-for-a-minute-no-one-will-notice-we-promise approach

Here is the race I am worried about with the above (which I'll rewrite slightly). 

```python
class Protocol(...):
  ...
  def __class_getitem__(cls, params):

    # Thread 1
    Protocol_old = typing.Protocol
    typing.Protocol = cls
                                                    # Thread 2
                                                    Protocol_old = typing.Protocol
                                                    typing.Protocol = cls
    try:
      return super().__class_getitem__(params)
    finally:
      typing.Protocol = Protocol_old
                                                    try:
                                                      return super().__class_getitem__(params)
                                                    finally:
                                                      typing.Protocol = Protocol_old
```

Consider:

``` python
import typing
from abc import abstractmethod
from functools import partial
from threading import Thread
from typing import TypeVar
from beartype.typing import Protocol

T = TypeVar("T")

contrived_example = """
class ThisIsNutsButMayBePossible{}(Protocol[T]):
  @abstractmethod
  def foo(self) -> None:
    pass
"""

def compileit(i: int):
    exec(contrived_example.format(i))

threads = [Thread(target=partial(compileit, i)) for i in range(100)]

for thread in threads:
    thread.start()

for thread in threads:
    thread.join()

print(typing.Protocol)  # <-- what result?
```

If I run that, I get the output `<class 'typing.Protocol'>` pretty consistently _until_ I add a `print()` after the `typing.Protocol = cls` line. _Then_ I get `<class 'beartype._protocol.Protocol'>`, also pretty consistently. Now I concede that this is contrived (yes, I know I'm deliberately triggering an interrupt with the artificial `print()` in the middle), and that this may be a non-existent risk in the real world.

## Yojimbo Baggins was right - revisiting creating and gutting the `_GenericAlias`

You are, as usual, completely justified in your suspicions. My use of `*gen_alias.__args__` was just flat out wrong. Further, as you point out, `_typevar_types` and `_paramspec_tvars` (introduced in Python 3.10) are MIA.

I thought of doing something like this:

``` python
        def __class_getitem__(cls, params):
            gen_alias = _Protocol.__class_getitem__(params).
            gen_alias.__origin__ = cls
            return gen_alias
```

But that is…_problematic_. From [`Protocol.__class_getitem`](https://github.com/python/cpython/blob/8b1b27f1939cc4060531d198fdb09242f247ca7c/Lib/typing.py#L1311-L1336):

``` python
    @_tp_cache  # <-- uh-oh
    def __class_getitem__(cls, params):
        …
        return _GenericAlias(cls, params,
                             _typevar_types=(TypeVar, ParamSpec),
                             _paramspec_tvars=True)
```

The value returned from `typing.Protocol.__class_getitem__(params)` is cached. If we change that value, later calls to `typing.Protocol.__class_getitem__` with the same `params` will get our hacked up version, _even if they are not from our our stand-in_. 🤦

[This](https://github.com/beartype/beartype/pull/86/files#r795035517) is the best alternative to swapping out `typing.Protocol` I can come up with.

If that meets not with your approval and that we should galaxy-swap away, just say the word, and it shall be done. I live but to serve. 🙇

[Galaxy Bear](https://www.deviantart.com/dimfann/art/Galaxy-Brain-820662977)

beartype/_protocol.py
```diff
@@ -0,0 +1,277 @@
...
```
posita, line 249: One last attempt….

posita: Hmmm…maybe not? `_typevar_types` might influence how `params` is interpreted on construction. That `typing.py` is full of some obfuscated stuff! Who can maintain anything like that?

Anyway, I think what we want is something like this:

``` python
if hasattr(fake_gen_alias, "_typevar_types"):
  new_gen_alias = type(fake_gen_alias)(cls, params,
    name=fake_gen_alias._name, inst=fake_gen_alias._inst,
    _typevar_types=fake_gen_alias._typevar_types,
    _paramspec_tvars=fake_gen_alias._paramspec_tvars)
else:
  new_gen_alias = type(fake_gen_alias)(cls, params,
    name=fake_gen_alias._name, inst=fake_gen_alias._inst)
```

TeamSpen210: To bypass the LRU cache, you could fetch `__wrapped__` on the func, which `@wraps` sets for you. Would that be easier?

posita: Yeah, I thought about that. I initially rejected it in case `@_tp_cache` ever went away, but I think that fear is pretty silly. I'm already up to my eyeballs in seven-layer-bean-and-private-implementation-details dip. Let me give it a shot. I think it will be cleaner.

beartype/_protocol.py
```diff
@@ -0,0 +1,269 @@
...
+            # gen_alias.__origin__ is (unsurprisingly) typing.Protocol, but we
+            # need it to be our class. Otherwise our inheritors end up with
+            # the wrong metaclass for some reason (i.e., type(typing.Protocol)
+            # instead of the desired _CachingProtocolMeta). Luddite alert: I
+            # don't fully understand the mechanics here. I suspect no one
+            # really does.
+            gen_alias.__origin__ = cls
+
+            # We're done! Time for a honey brewskie break. We earned it.
+            return gen_alias
```
posita, line 241: Second one more try….

beartype/_protocol.py
posita: Obsoleted. And wrong. So wrong.

beartype/typing.py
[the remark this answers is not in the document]
posita: I _think_ we decided to tackle caching later as needed (if it ever becomes an issue). Resolving for now, but we can revisit. If it's urgent for a release, I'll put that in a separate PR based on this one.

posita:
Rebased and added comment referencing https://bugs.python.org/issue46581, which would make my proposed approach for `__class_getitem__` more tenable. Perhaps we should take a dependency on that being fixed?

beartype/_protocol.py
```diff
@@ -0,0 +1,277 @@
...
+            # and we ain't in it. (RIP, GC.) Let's see if we can sneak in,
+            # shall we?
+
+            # FIXME: Once <https://bugs.python.org/issue46581> is addressed,
+            # consider replacing the madness below with something like:
+            #
+            #   cached_gen_alias = _Protocol.__class_getitem__(_Protocol, params)
+            #   our_gen_alias = cached_gen_alias.copy_with(params)
+            #   our_gen_alias.__origin__ = cls
+            #   return our_gen_alias
```
posita, line 223: This might be nice to wait for.

leycec:
> Now I concede that this is contrived (yes, I know I'm *deliberately* triggering an interrupt with the artificial `print()` in the middle), and that this may be a non-existent risk in the real world.

**Oh, no.** You're thoughtfully sane in a world full of dodgy characters with shifty unibrows and a knuckle-dragging gait like me. By inspection (*...so, wild and unsubstantiated assumption*), `typing.Generic.__class_getitem__()` doesn't appear to have any potential GIL releases – but that's never good enough, is it?

CPython releases the GIL in surprising circumstances. [See: "What is `print('')`, ghost of Trebek?"] Auditing the entire transitive stack of calls for a single GIL release across all Python versions present, future, and sidereal breaks my feeble grip on reality.

Likewise, permuting a copy-and-pasted version of `typing.Generic.__class_getitem__()` also obstructs sanity. That method has changed fairly radically since Python 3.8.

Luckily, you and @TeamSpen210 hashed out something subversive and incendiary behind everyone's unwashed backs – didn't you? The resulting implementation is a [surprisingly concise six-liner documented with a lucid foray into the sickly fever-dream that is the private `typing` API](https://github.com/beartype/beartype/pull/86/files#r796842129).

**I'm totally down with this entire wall of blasphemy.** The extensive working tests are only the gluten-free cheesecake topping the five-course cookie dough ice cream sundae of the meal. Still, I'd like to give everything a last once-over tomorrow with my "pedanto eye." <sup>...as my delightful wife calls it</sup>

## Future Work for a Future World

This is just an incidental note to myself and *not* at all an insinuation that anyone should do this for me here. Seriously, actually. This PR is stuffed to the gills and then forcefed to the Atlantic Bluefin Tuna.

That said:

* I dimly recall leveraging the stock `typing.Protocol` superclass on at least one prominent occasion in the codebase: `beartype/_util/hint/pep/proposal/utilpep544.py`, which internally maps from useless `typing.(Binary|Text|)IO` classes to our useful in-house analogues. For great speed and just because we can and should, that should be refactored in terms of this Nordic God-tier `beartype.typing.Protocol` superclass. That's *probably* a single-line switch out, but I assume nothing and expect even less.
* It'd be real swag to eventually generalize our existing PEP 544-specific test sample set at `beartype/_util/hint/pep/proposal/utilpep544.py` to test against both `typing.Protocol` *and* `beartype.typing.Protocol`. Since you've already implemented a parallel suite of compelling unit tests that guarantee goodness, "eventually" is the keyword of interest here.

## And... I'm Spent... Again.

In conclusion, this PR is literal:

> [HAPPINESS FOR EVERYBODY, FREE, AND NO ONE WILL GO AWAY UNSATISFIED!](http://soviethistory.msu.edu/wp-content/uploads/2015/10/picnic.pdf)

leycec:
**The review to end all reviews:** it's beginning.

popcorn please

leycec:
**Ohnoes.** All pending work for `beartype 0.10.0` has been shipped to the cleaners and is now waiting for me to just tag the most recent commit and ~~falsify up~~ accurately describe the changelog – which is *huge,* by the way. What a cray-cray release cycle that was, eh? Eh? <sup>← so much :canada: right there</sup>

Tragically, I took so long to review this that this *probably* won't land in `beartype 0.10.0`. I deeply and regretfully apologize to @posita and @TeamSpen210 with mournful tears streaming into my dinner, which now tastes disgustingly salty.

Thankfully, this is my top priority for the inevitable `beartype 0.10.1` point release. I *will* make this happen... or my name isn't Piston Sam! :steam_locomotive: :steam_locomotive: :steam_locomotive:

posita:
> Tragically, I took so long to review this that this _probably_ won't land in `beartype 0.10.0`. I deeply and regretfully apologize to @posita and @TeamSpen210 with mournful tears streaming into my dinner, which now tastes disgustingly salty.
> 
> Thankfully, this is my top priority for the inevitable `beartype 0.10.1` point release. I _will_ make this happen... or my name isn't Piston Sam! 🚂 🚂 🚂

I think the delay is good, and I wouldn't kill myself to squeeze this into 0.10.1. First, we might want to consider waiting until CPython gets is `copy_with` fix. (Of course, we might be waiting for awhile, so maybe that's a bad strategy.) Second, I think I'd feel comfortable with some time/space to test this out in some real world applications (which I haven't yet done to the extent I would like). Should I put this back in "draft" mode?

leycec:
> First, we might want to consider waiting until CPython gets its `copy_with` fix.

This is my preferred life strategy. Begging my betters to do things for me has yet to fail. B-b-ut...

> (Of course, we might be waiting for awhile, so maybe that's a bad strategy.)

**...this.** The CPython world moves like glacial meltwater off the Canadian permafrost infested with ancient virii first envisioned by John Carpenter.

if you burn it, it can die
<sup>*Python 3.10 release cycle dev struggle is real*</sup>

We'll be lucky to ram this thing through Python 3.11. Meanwhile, Python 3.7 through 3.10 still exist. (Python 3.6: *you are dead to me.*)

Thanks again for the tremendous work, everyone. I'm deathly afraid of breaking this PR and thus repressing my earnest desire to fundamentally refactor `beartype.typing` into something horrifying.

I... I don't know how long I can hold on, guys. :sweat:  :cold_sweat: :hot_face:

leycec:
@posita: On second thought, **this is good to go.** Don't sweat the overly fine minutiae and my banal code review. I'm all for banal, let's be clear – but you've already been around the bend (and back again) with @TeamSpen210.

I'll add a few internal `FIXME:` comments for one or both of us to *eventually* resolve these minor pending concerns, which can absolutely wait until then:

* The `if base is cls or base.__name__ in (...):` branch *probably* erroneously matches unrelated user-defined types whose names just happen to be `Generic` or `Protocol`. Ideally, we should tighten that up to only match the actual `{beartype,}.typing.{Generic,Protocol}` types.
* The call to `typing._get_protocol_attrs()` violates privacy encapsulation, which has me cueing up Megadeth's *Sweating Bullets* on the junkyard vinyl playlist. Ideally, we should copy-and-paste that function into this private submodule instead.

If heaven can wait, <sup>...Alexa says it can</sup> so can this.

Thunderous applause for this third out of three tremendous contributions, @posita. Likewise, thanks so much for the grand critique that helped fuel this along in my awkward and questionable absence, @TeamSpen210.

You both are the stuff open-source dreams are made of. :man_dancing:  :woman_dancing:

posita:
> * The `if base is cls or base.__name__ in (...):` branch _probably_ erroneously matches unrelated user-defined types whose names just happen to be `Generic` or `Protocol`. Ideally, we should tighten that up to only match the actual `{beartype,}.typing.{Generic,Protocol}` types.
> * The call to `typing._get_protocol_attrs()` violates privacy encapsulation, which has me cueing up Megadeth's _Sweating Bullets_ on the junkyard vinyl playlist. Ideally, we should copy-and-paste that function into this private submodule instead.

☝️ _This._ 

Your points are well taken. If I get some time to try to nudge things into a better place, I will (although `_get_protocol_attrs` is going to give me heartburn either way). Alternatively, if a `beartype` customer has problems with these, I will come to provide aid as soon as I can. I should be able to tackle the `base.__name__` issue shortly. The standard library makes a similar blunder in…_wait for it_…[`_get_protocol_attrs`](https://github.com/python/cpython/blob/278fdd3e3a2492665b2c2888fd2f428f7f59a3f5/Lib/typing.py#L1544-L1558), so maybe I can tackle both problems at once….

leycec:
> ...although `_get_protocol_attrs()` is going to give me heartburn either way...

Just got there fifteen minutes ago. Thanks to `typing`, I burn with the incandescent fury of a thousand suns imploding into the gaping maw of the Andromeda Galaxy.

**True worst-case end times story:** when [our galaxy collides with Andromeda in 5 billion years](https://en.wikipedia.org/wiki/Andromeda%E2%80%93Milky_Way_collision), armchair Wikipedia astrologists believe that "...**the Sun** might be brought near the centre of the combined galaxy, potentially coming near one of the black holes before being **ejected entirely out of the galaxy**.[[13]](https://en.wikipedia.org/wiki/Andromeda%E2%80%93Milky_Way_collision#cite_note-13) Alternatively, the Sun might approach one of the black holes a bit closer and be **torn apart by its gravity.** Parts of the former Sun would be **pulled into the black hole.**"

Is this supposed to reassure us, Wikipedia?

Are we just ping-pong balls to you, Wikipedia?

**Back to the issue front.** I realize I no longer understand the `_check_only_my_attrs()` helper. If we combine its powers with `_get_protocol_attrs()`, a terrifying new kaiju emerges from the radioactive dust plume:

``` python
    def _check_only_my_attrs(cls, inst: Any) -> bool:
        # So, "dict" actually supports set operators under
        # Python >= 3.9. Ignoring Python < 3.9 just cause,
        # this *probably* reduces to this one-liner:
        #     attrs = set(cls.__dict__ | cls.__dict__.get("__annotations__", {}))
        # We may not even need to coerce "attrs" to a set.
        # Dictionaries are now pretty much sets, but hotter.
        # *ANYWAY.* This makes sense; we're getting a
        # set of all unique attribute names. Good!
        attrs = set(cls.__dict__)
        attrs.update(cls.__dict__.get("__annotations__", {}))

        #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
        # Inlining _get_protocol_attrs(): it begins.
        #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

        # Uh... This is why we should inline. Let's pretend
        # this redefinition never happens. Eye is twitching.
        base_attrs = set()

        # *UH OH.* For all classes, the following tautology
        # holds:
        #     >>> cls.__mro__[0] is cls
        #     True
        # This means that the first "base" visited by this
        # iteration is just the passed "cls". But, wait. We
        # just introspected the attributes of "cls" above!
        # Clearly, we don't need to do that above. Let's
        # centralize all of the attribute inspection here.
        #
        # Very well. We progress deeper into the dimly
        # lit cavern from which grunting can be heard.
        for base in cls.__mro__[:-1]:  # without object
            # Guido, you cause me pain.
            if base.__name__ in ('Protocol', 'Generic'):
                continue

            # *WTFBRO.* What is this madness? Coercing
            # "KeyView" containers into lists and then
            # appending them into an even larger list?
            # This is a facepalm. This is why we inline.
            # Lastly, "{}" is not a singleton –
            # unlike "()", which is. Don't ask. So, we
            # declare a stupid empty global dictionary
            # singleton as a stupid hidden parameter: e.g.,
            #     def _check_only_my_attrs(
            #         cls, inst: Any, _empty_dict = {}) -> bool:
            #
            # This should then suffice with speed and glamour:
            #     base_dict = base.__dict__
            #     cls_attrs = base_dict | base_dict.get('__annotations__', _empty_dict)
            #     for attr in cls_attrs: ...
            #
            # Grunting sounds segue into mewling, hissing,
            # and a blood-curdling falsetto abruptly cut short.
            annotations = getattr(base, '__annotations__', {})
            for attr in list(base.__dict__.keys()) + list(annotations.keys()):
                # This is fine. It is the only thing that is.
                if not attr.startswith('_abc_') and attr not in EXCLUDED_ATTRIBUTES:
                    base_attrs.add(attr)

        #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
        # Inlining _get_protocol_attrs(): it ends.
        #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

        # I begin to question basic assumptions like
        # whether my 30 year-old username being a
        # portmanteau of leyline and Cecil is really
        # such a good idea after all.
        #
        # I've now officially lost the entire plot.
        # Doesn't this statement just reduce to:
        #     attrs = attrs
        # That is, doesn't intersection_update()
        # just remove all attribute names that are
        # *NOT* in both sets? But "base_attrs" is
        # the proper superset of "attrs". So, this
        # just removes all attribute names except
        # those directly defined by the passed "cls".
        # In other words, the entire doubly-nested
        # "for" loop we performed above was a noop.
        #
        # The sound of wet bloody meat being forcibly
        # dragged across a rough surface of stalagmites
        # percolates through the fetid crypt-like air.
        attrs.intersection_update(base_attrs)  # type: ignore [attr-defined]

        # And... we're iterating above. Pretty sure we can
        # just perform this test in the iteration above. That
        # is, I *would* be pretty sure if I was pretty sure
        # about anything here. But I'm not.
        #
        # Oppressive grunting sounds resume.
        for attr in attrs:
            if (
                not hasattr(inst, attr) or
                (
                    callable(getattr(cls, attr, None)) and
                    getattr(inst, attr) is None
                )
            ):
                return False

        return True
```

> The standard library makes a similar blunder in…wait for it…[`_get_protocol_attrs`](https://github.com/python/cpython/blob/278fdd3e3a2492665b2c2888fd2f428f7f59a3f5/Lib/typing.py#L1544-L1558).

Rictus grimace: *triggered.* :grimacing: 

Since you understandably find no deeper fulfilment in life than reporting CPython issues, <sup>i am this way too</sup> would you mind taking one for all humanity by throwing yourself onto the sword that is the public CPython bug tracker? @TeamSpen210 and I loudly cheer from the sidelines at a distance.

It's a disturbing shocker that this is still in CPython 3.10. The tenuous compassion for all mankind I (of course) normally feel has slipped yet another femptometer closer to the abyssal cliff of Nietzschean despair.

abyss is cat

leycec:
*ohisortagetitnow*

You... you are actually bloody ingenious, @posita. I had to judiciously pollute your `_check_only_my_attrs()` function with untold `print()` statements to even begin to comprehend the dark mysteries unfolding before me as I slumped over the keyboard.

The ominous key appears to be that CPython implicitly calls the metaclass `__instancecheck__()` method *once for each subclass* of a `beartype.typing.Protocol` hierarchy – each of which necessarily shares your same caching metaclass and thus that method. I think? If I'm reading these `print()` statements correctly? I'm probably not.

That's fascinating if true. I'd assumed CPython would only call `__instancecheck__()` *once for the entire hierarchy,* which only goes to show that I resemble a donkey. My wife agrees.

I still have no clear idea what exactly this does, though:

```python
        attrs.intersection_update(_get_protocol_attrs(cls))  # type: ignore [attr-defined]
```

Clearly, that's essential. I know this because the Abyss opens up when I comment that out. But I'm also clearly incapable of fully comprehending the eternal verities that you have authored here. Please enlighten me, Beneficent Protocol Guru, that I may comment this and then inline the body of `_get_protocol_attrs()` here forthwith. :mage_man:

leycec:
**Hah!** We see you deleting branches over there. (≖_≖ )

leycec:
**Buh-yah.** Finally untangled the rat's nest. Behold a type-checking subroutine for caching protocols respectful of privacy (and faster on that first uncached hit, too):

```python
    def _check_only_my_attrs(cls, inst: Any, _EMPTY_DICT = {}) -> bool:

        cls_attr_name_to_value = cls.__dict__
        cls_attr_name_to_hint = cls_attr_name_to_value.get(
            '__annotations__', _EMPTY_DICT)
        cls_attr_names = (
            cls_attr_name_to_value | cls_attr_name_to_hint
            if IS_PYTHON_AT_LEAST_3_9 else
            dict(cls_attr_name_to_value, **cls_attr_name_to_hint)
        )

        # For the name of each attribute declared by this protocol class...
        for cls_attr_name in cls_attr_names:
            # If...
            if (
                # This name implies this attribute to be unignorable *AND*...
                #
                # Specifically, if this name is neither...
                not (
                    # A private attribute specific to dark machinery defined by
                    # the "ABCMeta" metaclass for abstract base classes *OR*...
                    cls_attr_name.startswith('_abc_') or
                    # That of an ignorable non-protocol attribute...
                    cls_attr_name in _PROTOCOL_ATTR_NAMES_IGNORABLE
                # This attribute is either...
                ) and (
                    # Undefined by the passed object *OR*...
                    #
                    # This method has been specifically "blocked" (i.e.,
                    # ignored) by the passed object from being type-checked as
                    # part of this protocol. For unknown and presumably
                    # indefensible reasons, PEP 544 explicitly supports a
                    # fragile, unreadable, and error-prone idiom enabling
                    # objects to leave methods "undefined." What this madness!
                    not hasattr(inst, cls_attr_name) or
                    (
                        # A callable *AND*...
                        callable(getattr(cls, cls_attr_name, None)) and
                        # The passed object nullified this method. *facepalm*
                        getattr(inst, cls_attr_name) is None
                    )
                )
            ):
                # Then the passed object violates this protocol. In this case,
                # return false.
                return False

        # Else, the passed object satisfies this protocol. In this case, return
        # true.
        return True
```

@posita: **Uhohs.** Turns out caching protocols fail to support a standard use case of subscription by non-type variables satisfying a type variable: e.g.,

```python
# This is fine.
>>> from typing import AnyStr, Protocol
>>> class GoodProtocol(Protocol[AnyStr]): pass
>>> class BadProtocol(GoodProtocol[str]): pass

# This blows fine chunks across my keyboard.
>>> from beartype.typing import AnyStr, Protocol
>>> class GoodProtocol(Protocol[AnyStr]): pass
>>> class BadProtocol(GoodProtocol[str]): pass
  File "mopy.py", line 7, in <module>
    class BadProtocol(GoodProtocol[str]): pass
  File "/home/leycec/py/beartype/beartype/typing/_typingpep544.py", line 403, in __class_getitem__
    gen_alias = _Protocol.__class_getitem__.__wrapped__(
  File "/usr/lib/python3.8/typing.py", line 890, in __class_getitem__
    raise TypeError(
TypeError: Parameters to Protocol[...] must all be type variables
```

Do `numerary` protocols suffer a similar issue? You know it's Wednesday when you feel like this. :face_exhaling:

posita:
Wow…that is some heavy lifting! I hope you did not suffer permanent injury.

My apologies for my lack of response on this thread. (Pesky day job has me runnin' wild.) I promise to circle back as soon as I get a minute! 😅

P.S. Don't worry about branches. New ones are easy to make. 😉

leycec:
Oh, no worries whatsoever! Work-life balance is no balance when you're teetering on the edge. I just accidentally resolved the above issue, too. I don't entirely understand the solution, which is par for the course around here:

``` python
        # Newer, better, faster, harder __class_getitem__()
        # builds a brighter world despite watering eyes and
        # increased sanity loss.
        def __class_getitem__(cls, item):

            # We have to redefine this method because typing.Protocol's version
            # is very persnickety about only working for typing.Generic and
            # typing.Protocol. That's an exclusive club, and we ain't in it.
            # (RIP, GC.) Let's see if we can sneak in, shall we?

            # FIXME: Once <https://bugs.python.org/issue46581> is addressed,
            # consider replacing the madness below with something like:
            #   cached_gen_alias = _Protocol.__class_getitem__(_Protocol, params)
            #   our_gen_alias = cached_gen_alias.copy_with(params)
            #   our_gen_alias.__origin__ = cls
            #   return our_gen_alias

            # We can call typing.Protocol's implementation directly to get the
            # resulting generic alias. We need to bypass any memoization cache
            # to ensure the object on which we're about to perform surgery
            # isn't visible to anyone but us.
            if hasattr(_Protocol.__class_getitem__, '__wrapped__'):
                # !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
                # @posita: New magic happens here.
                # !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
                base_cls = (_Protocol if _Protocol in cls.__bases__ else cls)
                gen_alias = super().__class_getitem__.__wrapped__(
                    base_cls, item)
            else:
                # We shouldn't ever be here, but if we are, we're making the
                # assumption that typing.Protocol.__class_getitem__ no longer
                # caches. Heaven help us if it ever uses some proprietary
                # memoization implementation we can't see anymore because it's
                # not based on functools.wraps.
                gen_alias = _Protocol.__class_getitem__(item)

            # Now perform origin-replacement surgery. As-created,
            # gen_alias.__origin__ is (unsurprisingly) typing.Protocol, but we
            # need it to be our class. Otherwise our inheritors end up with
            # the wrong metaclass for some reason (i.e., type(typing.Protocol)
            # instead of the desired _CachingProtocolMeta). Luddite alert: I
            # don't fully understand the mechanics here. I suspect no one does.
            gen_alias.__origin__ = cls

            # We're done! Time for a honey brewskie break. We earned it.
            return gen_alias
```

Because this makes my eyes water, this might make your eyes water, too. I recant everything I said about @beartype and `numerary` "going their own ways" on caching protocol implementations.

Would you like to unify our two approaches? As I dimly recall, our current caching protocol API *almost* satisfied your use cases – except for `include()` and `exclude()` methods, I think? If so, perhaps we might either add those methods to our API *or* enable `numerary` to safely subclass, monkey-patch, or otherwise add those methods to our API.

Because you're in the Bear Clan, you now have full *carte blanche* to:

* Add anything you like to `beartype.typing.Protocol`, even if `numerary` is the only thing that needs those things. *Things: lotsa things.*
* Publicize any currently private attributes of our caching protocol API, especially `_CachingProtocolMeta` (if you need it, of course).

It's a full moon out there and we're howling at it. :full_moon_with_face:

leycec:
For better or worse, this just got pushed out the door when no one was looking with `beartype 0.10.1` thanks to [a critical defect with respect to positional-only parameters](https://github.com/beartype/beartype/issues/101). Will `beartype.typing.Protocol` actually behave itself or am I sitting all alone on another miswired timebomb? :bomb: :boom: 

**You decide:**

```bash
$ pip install --upgrade beartype
```

posita:
Catching up on this…do I understand from https://github.com/beartype/beartype/blame/af566ed25b6fdc3dc89fdb5e2f8bf976f80e04e8/beartype/typing/_typingpep544.py#L382-L464 that the approach you outlined in https://github.com/beartype/beartype/pull/86#issuecomment-1043513088 has made it in? I _think_ that's the case, but I want to be sure I'm responding to the right thing….

leycec:
> Do I understand from [BIGLY URL] that the approach you outlined in [OTHER BIGLY URL] has made it into main (and also `0.10.1`)?

You understand everything. Thus, you understand this.

Actually, I'm jelly at how fast you grok code even in the midst of a feverish working pitch, full-throttle family fun, and open-source commitments beyond even my most wide-eyed idealism.

You dah bro. That's what I'm saying here.

> `numerary` loses 3.7 support in the process, but I think that's [worth it](https://github.com/posita/numerary/blob/posita/1/isolate-protocol-impl/docs/notes.md#040).

:face_with_head_bandage: 

On the bright side, [Python 3.7 hits EOL in a year-and-a-half]( https://endoflife.date/python). On the dark side, something worked *and now it doesn't.*

I'm actually curious how you achieved that when... **oh. That is right.** I recall it like it was yesterday's fetid goat cheese. The private `typing._Protocol` class in Python 3.7, right? Is that class actually a perfectly working drop-in replacement for `typing.Protocol` in Python ≥ 3.8? If so:

* **That's crazy.** Why hide beautiful things, Guido?
* **Let's use that.** Specifically, let's just substitute `typing._Protocol` for `typing.Protocol` like you probably currently do in `numerary`. Would that suffice?

It goes without saying, but I'll say it: "Please feel free to have a go at everything in the `beartype.typing` subpackage. *Mi casa es su casa,* because we built this house together. Okay, it was mostly you."

Relatedly:

> This decision was not made lightly. `numerary` is intended as a temporary work-around. It’s obsolescence will be something to celebrate.

:partying_face: 

*Wait just a hot minute!* That's not something to celebrate. `numerary` rocks. Since `numerary` rocks and you intend to quietly axe it out back, would you prefer to just shift its intestines into an appropriate @beartype subpackage? For example:

* **`beartype.numbers`.** Right? Right? Hear me out here. `numerary` is to [the standard `numbers` module](https://docs.python.org/3/library/numbers.html) as `beartype.typing` is to the standard `typing` module – sort of. Okay, not really. But let's just roll with this awkward analogy. So, `beartype.numbers` would basically be the supercharged runtime-friendly numeric typing API that everyone always wanted but only @posita was brave enough to chain his immaculate reputation, GitHub karma, and lasting legacy to.
* Alternately, **just stuff everything into `beartype.typing`.** I'd kinda prefer to maintain a one-to-one mapping with `typing`, but I can also see the argument for a monolithic typing API that *just does everything.*

You make the call. I'll roll the dice and tackle the issue tracker uproar.

PR #88, opened 2022-01-27
Adds beartype-generated lines (no cash) to linecache

### LINKED ISSUE #85: [Feature Request] Debuggable wrapper functions
`@beartype` dynamically wraps each decorated callable with a unique wrapper function efficiently type-checking that callable. That's great, because efficient. But that's also currently invisible to debuggers, because `@beartype` fails to register the Python code underlying these wrapper functions with [the little-known standard `linecache` module](https://docs.python.org/3/library/linecache.html).

This feature request tracks work towards rendering `@beartype` compatible with debuggers. For now, please refer to [this superb writeup by witty QA strongman @posita for a detailed discussion of the exact issue under discussion here – complete with the inevitable working solution infused with wisdom by cautionary AI futurologist @TeamSpen210](https://github.com/beartype/beartype/discussions/84).

Humble gratitude to both @posita and @TeamSpen210 for finally shaming me into doing this thing that desperately needs doing.

posita:
Resolve #85.

Allows for debugger visibility in many contexts. (This won't work in VSCode _[by default](https://code.visualstudio.com/docs/python/debugging#_justmycode)_, but that's [on Microsoft](https://github.com/microsoft/debugpy/issues/828), not us.) @TeamSpen210 did the heavy lifting on this (see #84). I just typed it in.

### DIFF
--- a/beartype/_util/func/utilfuncmake.py
+++ b/beartype/_util/func/utilfuncmake.py
@@ -20,6 +20,8 @@ from beartype._util.text.utiltextmunge import number_lines
 from collections.abc import Callable
 from functools import update_wrapper
 from typing import Optional, Type
+from weakref import finalize
+import linecache
 
 # See the "beartype.cave" submodule for further commentary.
 __all__ = ['STAR_IMPORTS_CONSIDERED_HARMFUL']
@@ -152,33 +154,29 @@ def make_func(
         # Python fails to capture that function (i.e., expose that function to
         # this function) when the locals() dictionary is passed; instead, a
         # unique local dictionary *MUST* be passed.
-        #
-        # Note that the same result may also be achieved via the compile()
-        # builtin and "types.FunctionType" class: e.g.,
-        #     func_code_compiled = compile(
-        #         func_code, "<string>", "exec").co_consts[0]
-        #     return types.FunctionType(
-        #         code=func_code_compiled,
-        #         globals=_GLOBAL_ATTRS,
-        #         argdefs=('__beartype_func', func)
-        #     )
-        #
-        # Since doing so is both more verbose and obfuscatory for no tangible
-        # gain, the current circumspect approach is preferred.
-        exec(func_code, func_globals, func_locals)
 
-        #FIXME: See above.
-        #FIXME: Should "exec" be "single" instead? Does it matter? Is there any
-        #performance gap between the two?
-        # func_code_compiled = compile(
-        #     func_code, func_wrapper_filename, "exec").co_consts[0]
-        # return FunctionType(
-        #     code=func_code_compiled,
-        #     globals=_GLOBAL_ATTRS,
-        #
-        #     #FIXME: This really doesn't seem right, but... *shrug*
-        #     argdefs=tuple(local_attrs.values()),
-        # )
+        # Make up a filename for compilation and possibly the linecache entry
+        # (if we make one). A fully-qualified name and ID *should* be unique
+        # for the life of the process.
+        func_full_name = (
+            f"{func_wrapped.__module__}{func_wrapped.__name__}"
+            if func_wrapped else
+            func_name
+        )
+        linecache_file_name = f"<@beartype({func_full_name}) at {id(func_wrapped):#x}>"
+
+        # We use the more verbose and obfuscatory compile() builtin instead of
+        # simply calling exec(func_code, func_globals, func_locals) because
+        # exec does not provide a way to set the resulting function object's
+        # .__code__.co_filename read-only attribute. We can use "single"
+        # instead of "exec" if we are willing to accept that func_code is
+        # constrained to a single statement. In casual testing, there is very
+        # little performance difference (with an imperceptibly slight edge
+        # going to "single").
+        func_code_compiled = compile(
+            func_code, linecache_file_name, "exec")
+        assert func_name not in func_locals
+        exec(func_code_compiled, func_globals, func_locals)
     # If doing so fails for any reason...
     except Exception as exception:
         # Raise an exception suffixed by that function's declaration such that
@@ -234,6 +232,23 @@ def make_func(
         # Document that function.
         func.__doc__ = func_doc
     # Else, that function is undocumented.
+
+    # Since we went through the trouble of printing its definition, we might
+    # as well make its compiled version debuggable, too.
+    if is_debug:
+        linecache.cache[linecache_file_name] = (  # type: ignore[assignment]
+            len(func_code),  # type: ignore[assignment]  # Y u gotta b diff'rnt Python 3.7? WHY?!
+            None,  # mtime, but should be None to avoid being discarded
+            func_code.splitlines(keepends=True),
+            linecache_file_name,
+        )
+
+        # Define and register a cleanup callback for removing func's linecache
+        # entry if func is ever garbage collected.
+        def remove_linecache_entry_for_func():
+            linecache.cache.pop(linecache_file_name, None)
+
+        finalize(func, remove_linecache_entry_for_func)
 
     # Return that function.
     return func
--- a/beartype_test/a00_unit/a90_decor/test_decorconf.py
+++ b/beartype_test/a00_unit/a90_decor/test_decorconf.py
@@ -118,6 +118,52 @@ def test_decor_conf_is_debug(capsys) -> None:
     assert '# is <function _earthquake ' in standard_captured.out
 
 
+def test_decor_conf_is_debug_updates_linecache(capsys) -> None:
+    '''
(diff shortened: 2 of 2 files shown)


posita:
@TeamSpen210 did the heavy lifting here (see #84). I just typed it in.

So this is neat:

``` python
# test.py
from beartype import beartype, BeartypeConf

@beartype(conf=BeartypeConf(is_print_wrapper_code=True))
def foo(i: int) -> str:
    breakpoint()
    return str(i)

print(foo(0))
```

``` sh
python test.py
(line 0001) def foo(
(line 0002)     *args,
(line 0003)     __beartype_func=__beartype_func, # is <function foo at 0x106c249d0>
…
(line 0041)             )
(line 0042)
(line 0043)     return __beartype_pith_0
> /…/test.py(7)foo()
-> return str(i)
(Pdb) bt
  /…/test.py(9)<module>()
-> print(foo(0))
  <@beartype(foo) at 0x106c249d0>(29)foo()
-> __beartype_pith_0 = __beartype_func(*args, **kwargs)
> /…/test.py(7)foo()
-> return str(i)
(Pdb) u
> <@beartype(foo) at 0x106c249d0>(29)foo()
-> __beartype_pith_0 = __beartype_func(*args, **kwargs)
(Pdb) l
 24                     pith_value=__beartype_pith_0,
 25                 )
 26
 27         # Call this function with all passed parameters and localize the value
 28         # returned from this call.
 29  ->     __beartype_pith_0 = __beartype_func(*args, **kwargs)
 30
 31         # Noop required to artificially increase indentation level. Note that
 32         # CPython implicitly optimizes this conditional away. Isn't that nice?
 33         if True:
 34             # Type-check this passed parameter or return value against this
(Pdb) c
0
```

posita:
I have no idea what [that error](https://github.com/beartype/beartype/runs/4968911087?check_suite_focus=true) is about. 😕

leycec:
**This so hot.** Plentiful bear hugs to the crack commando squad that made this happen. What cannot @posita and @TeamSpen210 do when their formidable powers combine? We are now finding out.

I'd better merge this ASAP before I accidentally break everything by drunkenly stumbling back into the `utilfuncmake` submodule on a frostbitten Saturday evening. Three questions for the restless crowd angrily thumping the armrests of their La-Z-Boys:

* The option name `is_print_wrapper_code`. It is verbose, overly specific, and I already grow weary of seeing it everywhere. But laziness prevails, because I else have yet to release `beartype 0.10.0` or publicly document anything. This PR generalizes that option to *also* cache that code for debuggability – which is great, because debuggability was the entire point of that option. Right? You are now wondering: "Where is the question we were promised." Okay, already! Down, angry rabble! Down! The question is:
  * Should we just name this option `is_debug` instead? My initial reasoning for that name was that we'd have different kinds of debugging options – like, say, `is_log_wrapper_code` or `is_linecache_wrapper_code`. That now seems like gross overkill. If anyone furiously demands deeper debugging options, we can always define those later. So... `is_debug`? Thumbs up? Facepalm? Double facepalm, even? :grimacing: 
* Kinda afraid to ask what's going on down below, but gonna ask anyway. What's going on down below? ↓
``` python
        finalize(func, lambda: linecache.cache.pop(linecache_file_name, None))
```
* This seem kinda sus and I am squinting af:

``` python
       linecache_file_name = f"<@beartype({func_name}) at {id(func_wrapped):#x}>"
       ...
 
       # FIXME: Do we need a loop here to avoid name clashes (a la
        # _make_method in
        # <https://github.com/python-attrs/attrs/blob/main/src/attr/_make.py>)?
        # Probobably unlikely, but there is this (from
        # <https://docs.python.org/3/library/functions.html#id>): "Two objects
        # with non-overlapping lifetimes may have the same id() value."
        linecache.cache[linecache_file_name] = (  # type: ignore[assignment]
```

As you astutely note, object IDs are technically ambiguous across the lifecycle of a Python process. Since we can't disambiguate on that, let's disambiguate on the `func_name` instead! Specifically, let's conditionally expand the assignment of `linecache_file_name` to prefer a fully-qualified function name cobbled together from the passed wrapper function if passed: e.g.,

``` python
       func_full_name = (
           f'{func_wrapped.__module__}{func_wrapped.__name__}
           if func_wrapped else
           func_name
       )
       linecache_file_name = f"<@beartype({func_full_name}) at {id(func_wrapped):#x}>"
```

Since `@beartype` always passes the currently decorated callable as the `func_wrapped` parameter to `make_func`, this *should* have the intended effect of disambiguating cache entries on fully-qualified callable names – which *should* be fully disambiguous across the lifecycle of a Python process (assuming no shenanigans, because let's not even go there, because my forehead capillary is already bursting out).

posita:
## On `is_debug`

> * Should we just name this option `is_debug` instead? My initial reasoning for that name was that we'd have different kinds of debugging options – like, say, `is_log_wrapper_code` or `is_linecache_wrapper_code`. That now seems like gross overkill. If anyone furiously demands deeper debugging options, we can always define those later. So... `is_debug`? Thumbs up? Facepalm? Double facepalm, even? 😬

My vote is double secret probation triple-lindy thumbs up. #90 is up to see what that looks like. If that works for you, I can rebase this PR onto that one.

## On `finalize`

> * Kinda afraid to ask what's going on down below, but gonna ask anyway. What's going on down below? ↓
> 
> ```python
>         finalize(func, lambda: linecache.cache.pop(linecache_file_name, None))
> ```

You are right to inquire. Your allusion to La-Z-Boys could describe me in ~writing that~ _obfuscating what was going on there_. I have attempted to make it a bit more accessible. I'll reproduce that attempt here:

``` python
        # Define and register a cleanup callback for removing func's linecache
        # entry if func is ever garbage collected.
        def remove_linecache_entry_for_func():
            linecache.cache.pop(linecache_file_name, None)

        finalize(func, remove_linecache_entry_for_func)
```

## Regarding uniqueness of file names

I like your proposal and adopted it with vigor (and added a comment).

## Side trip: Should "exec" be "single" instead? Does it matter?

Regarding this…

``` python
        #FIXME: Should "exec" be "single" instead? Does it matter? Is there any
        #performance gap between the two?
```

While I was messing with all of this, I figured I'd try to answer that because I was curious myself. In my superficial testing, `"exec"` is at worst about 1 µs slower than `"single"`, so not really a performance issue.

Here's my IPython test script:

``` python
# foo.ipy
code = """
class Foo:
  def __init__(self):
    self._foo = foo(1)
  @property
  def foo(self) -> int:
    return int(self._foo)
"""
%timeit compile(code, "foo.py", "exec")
%timeit compile(code, "foo.py", "single")
```

These are typical for my laptop:

``` sh
% ipython foo.ipy
38.9 µs ± 307 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
37.4 µs ± 444 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
% ipython foo.ipy
38.5 µs ± 454 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
39 µs ± 1.32 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
% ipython foo.ipy
40.6 µs ± 1.39 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
39.8 µs ± 1.13 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

The only other reason I could think of to use `"single"` is that it acts as a "safeguard" of sorts by limiting the generated code to defining a single thing. For example, `"single"` will fail on the following, where `"exec"` won't:

``` python
class Foo:
  def __init__(self):
    self._foo = foo(1)
  @property
  def foo(self) -> int:
    return int(self._foo)
def foo(a: int) -> str:  # <-- "single" will see this second thing as a syntax error
  return str(a)
```

I removed your `FIXME` and kept the `"exec"`, but let me know if I should change it to `"single"` and/or add a comment explaining any reasoning.

TeamSpen210:
`weakref.finalize()` is a tool which uses the weakref system to effectively add on a `__del__()` method to an object - it stores a callable, when the object dies it gets called. So this ensures if the original function is deleted (say you're beartyping a closure, perhaps?) we go and try to clear the linecache entry. `pop`'s desirable to ensure if someone else cleared the cache or something it doesn't error.

Ah, ninjaed.

posita:
Rebased in light of #90 landing.

leycec:
the impossible made possible

When a brilliance is so bright that it caustically sears itself into the mucosal lining of the back of your retinas, it's a safe bet @posita and @TeamSpen210 teamed up yet again to radically transform our humble backwater world into the utopian Niven ring of tomorrow, today.

Thanks *sooo* much for the tremendous contribution. Hot off the presses:

> **Crack :bear: Squad Crushes Another Tough Nut.**

PR #90, opened 2022-01-29
Rename is_print_...code to is_debug

posita:
Raised [here](https://github.com/beartype/beartype/pull/88#issuecomment-1024866336).


leycec:
**Holly balls of Batman.** When @posita executes the job, he does so swiftly and dispassionately – without remorse and at the Corinthian height of his prowess. Chittering bugs that are rapidly multiplying out of control:

> *you are on notice.*

posita:
Wink, finger gun.

Issue #109, opened 2022-03-05
Ruh-roh! CACHING protocols? More like CRASHING protocols (in 3.7)! Amirite?


posita:
So here's a head-scratcher:

``` python
# test_ruh_roh.py
from abc import abstractmethod
from beartype import beartype
from beartype.typing import TYPE_CHECKING, Iterable, Protocol, TypeVar, Union, runtime_checkable

_T_co = TypeVar("_T_co", covariant=True)

if TYPE_CHECKING:
    from abc import ABCMeta as _BeartypeCachingProtocolMeta
else:
    _BeartypeCachingProtocolMeta = type(Protocol)

# - - - - - - - - - - - - - - - - SupportsInt - - - - - - - - - - - - - - - -

@runtime_checkable
class SupportsInt(
    Protocol,  # <-- this is totally cool
    metaclass=_BeartypeCachingProtocolMeta,
):
    __slots__: Union[str, Iterable[str]] = ()
    @abstractmethod
    def __int__(self) -> int:
        pass

@beartype  # <-- this works
def myint(arg: SupportsInt) -> int:
    return int(arg)

print(f"{myint(0.0)} == {0}?")
assert myint(0.0) == 0

# - - - - - - - - - - - - - - - - SupportsAbs[_T_co] - - - - - - - - - - - - - - - -

@runtime_checkable
class SupportsAbs(
    Protocol[_T_co],  # <-- uh-oh, now we're askin' for trouble
    metaclass=_BeartypeCachingProtocolMeta,
):
    __slots__: Union[str, Iterable[str]] = ()
    @abstractmethod
    def __abs__(self) -> _T_co:
        pass

@beartype  # <-- boom
def myabs(arg: SupportsAbs[_T_co]) -> _T_co:
    return abs(arg)

print(f"{myabs(-1)} == {1}?")
assert myabs(-1) == 1
```

Using beartype @ 8da3ab48e80d37267b38938f22bf8ce2d1c070ba in 3.8:

``` sh
% python3.8 -m mypy test_ruh_roh.py
Success: no issues found in 1 source file
% python3.8 test_ruh_roh.py
0 == 0?
1 == 1?
```

And now 3.7:

``` sh
% python3.7 -m mypy test_ruh_roh.py
Success: no issues found in 1 source file
% python3.7 test_ruh_roh.py
0 == 0?
Traceback (most recent call last):
  File "test_ruh_roh.py", line 45, in <module>
[...]

PR #110, opened 2022-03-05
[WIP] Failing test for #109

posita:
It's a start.


leycec:
Work in progress: **your work ends here.** This is actually excellent as is. Unit tests exhibiting shameful failures are incredibly useful. I'm considering hacking away at the gangrenous limb that is #109.

Merging this helps make that happen. We doin' this! :running_man:

PR #118, opened 2022-04-05
Make protocol `__slots__` more type-"friendly" (i.e., work around mypy bug)

posita:
See python/mypy#11013 for details, but we may impose downstream heartache on those
deriving from our protocols if we don't explicitly set the type of `__slots__` to `Any`.

This *also* corrects for my previously botched metaclass TypeVar in accordance with
[this recommendation](https://github.com/python/mypy/issues/9282#issuecomment-1082483033=).
Could I have included that fix as a separate PR? Yes. Should I have? Yes. Did I? No.
`#GottaGoFast` or something the kids say. (Translation: Because I'm lazy.)

### DIFF
--- a/beartype/typing/_typingpep544.py
+++ b/beartype/typing/_typingpep544.py
@@ -47,7 +47,6 @@ if IS_PYTHON_AT_LEAST_3_8:
         SupportsInt as _SupportsIntSlow,
         SupportsRound as _SupportsRoundSlow,
         TypeVar,
-        Union,
         runtime_checkable,
     )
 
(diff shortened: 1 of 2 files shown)


leycec:
**Wondrous breath of Spring!** It's equally disheartening and unsurprising that mypy prefers the meaningless catchall `Any` to the meaningfully disambiguous `Union[str, Iterable[str]]`.

...but that's mypy in a mouldy nutshell. When you're the big guy with the improvised nail bat, you don't have to make sense; you just have to improve morale with omnipresent beatings.

mypy reasoning

> This also corrects for my previously botched metaclass TypeVar in accordance with https://github.com/python/mypy/issues/9282#issuecomment-1082483033=.

**Blessed be,** it's @posita with a two-for-one special.

Since this is unilateral brilliance, it's obviously in. Also, this is probably obvious, but:

* You know exactly what you are doing.
* I unconditionally trust you. I don't even unconditionally trust our Bengal cat. Admittedly, that's because she routinely claws suppurating gouges into my once-dextrous fingers.
* Submitting PRs consumes valuable time away from your delightful family, competitive board game group that frighten neighbours, and growing retinue of projects dominating all the GitHub charts.

What I'm saying is... I *love* hearing from you, but I also acknowledge that sometimes you just wanna push without subjecting yourself to the labyrinthian gauntlet of PR pain.

PR #136, opened 2022-06-04
feat: adding math.TypeHint and is_subclass

### LINKED ISSUE #133: [Feature Request] New `beartype.math` API for performing type hint arithmetic
### Discussed in https://github.com/beartype/beartype/discussions/132

<div type='discussions-op-text'>

<sup>Originally posted by **tlambert03** May 31, 2022</sup>
First, thanks for this amazing library.  The scope and detail are beary impressive. 🐻

Apologies if I've missed an obvious answer (or previous discussion) of this question: I've been digging through the readme and source code for a couple days but still haven't hit on it, and I'm not sure if this is just fundamentally out of the primary goal here of `isinstance(some_object, some_typehint)`.

I'm looking for something that would tell me whether type hint B is "compatible" with type hint A (something akin to pytypes `is_subtype` function):

```python
from typing import Sequence, List, Union, Optional, Any

def is_subtype(hintA, hintB) -> bool:
    ...  # ?

assert is_subtype(list, list) is True
assert is_subtype(list, Sequence) is True
assert is_subtype(Sequence, list) is False

assert is_subtype(List[int], Sequence[Any]) is True
assert is_subtype(List[int], Sequence[int]) is True
assert is_subtype(List[int], Sequence[str]) is False

assert is_subtype(list, Union[list, str]) is True
assert is_subtype(Union[int, str], Union[int, str, list]) is True
assert is_subtype(Union[list, str], list) is False
assert is_subtype(list, Union[int, str]) is False
assert is_subtype(Union[int, tuple], Union[int, str, list]) is False

assert is_subtype(Optional[int], int) is False
assert is_subtype(int, Optional[int]) is True
```

is this something that could be achieved with `beartype`?  digging through the source, i see _lots_ of stuff that seems very useful towards this goal, but not sure I see anything in the public API (since we're not really trying to determine whether some object is an instance of a hint).

my goal here would be to detect if an API breakage has occurred between two versions:

```py
# version 1
def func() -> list:
   ...
```

```py
# version 2
def func() -> Sequence:  # potential API breakage
   ...
```

... and this seems like something you'd have an opinion on :) 
thanks for your work!
</div>

tlambert03:
closes #133

lots of things to be cleaned-up, documented, and discussed here.  But opening a PR to discuss progress on this.

You can see currently passing checks in test_is_subtype (not sure where I should be putting the test in the suite).

Will ping you with more specific questions/topics when ready


leycec:
**Brilliance personified.** I *cannot* believe you bit-banged that out in a night. But, Harvard, so... I guess I can believe. You amaze even as you stun, @tlambert03.

Welp. Let's get this `git` party started. I'll probably subject all of us to a mind-numbing peer review that makes us question the wisdom of the open-source crowd-sourced model – at some point. Because this is Friday night, this isn't that point. Rejoice! Instead...

> You can see currently passing checks in `test_is_subtype`

:open_mouth: 

> (not sure where I should be putting the test in the suite).

...*heh.* Yeah. So, our test suite is obsessive-compulsively organized to guarantee that unit tests exercising lower-level functionality are run *before* unit tests exercising higher-level functionality. Wouldn't want to test the public `@beartype` decorator itself before first testing that all of the private utility functions inside the `beartype._util` subpackage called by that decorator behave as expected, right?

Thankfully, nothing and nobody else actually use the `beartype.math` subpackage. <sup>*...yet*</sup>  It's actually fine for now to just stuff the `test_is_subtype` submodule directly into the `beartype_test` package – exactly as you've wisely done.

Eventually, a reasonable landing place for that submodule would probably be within a new test subpackage living at:

```
beartype_test.a00_unit.a60_api.math.test_is_subtype
```

Dissecting that:

* `a00_unit` forces lower-level unit tests to run *before* the higher-level "functional" integration tests in `a90_func`.
* `a60_api` forces mid-level API unit tests to run *after* both the lower-level utility tests in `a20_util` and PEP compliance tests in `a40_pep` but *before* both the higher-level error-handling tests in `a80_error` and decoration tests in `a90_decor`.

Unrelatedly, I now sadly note that...

## Type Hints Are Only Partially Ordered

**Ohnoes!** I committed the cardinal sin of wishful thinking when I applied the `@functools.total_ordering` decorator to the `beartype.math.TypeHint` class.

Type hints *almost* constitute a total ordering under subtype compatibility; they satisfy reflexivity, transitivity, and antisymmetry as expected. Tragically for us, they violate totality (i.e., [strongly connected](https://en.wikipedia.org/wiki/Connected_relation)). Bleakly consider:

```python
>>> from collections.abc import Awaitable
>>> TypeHint(list[int]) <= TypeHint(Awaitable[str])
False
>>> TypeHint(Awaitable[str]) <= TypeHint(list[int])
False
```

Right? Okay. So that means type hints violate totality. If type hints satisfied totality, then either `TypeHint(list[int]) <= TypeHint(Awaitable[str])` *or* `TypeHint(Awaitable[str]) <= TypeHint(list[int])` would be `True`. But both conditions are `False`, because a container type hint like `list[int]` and a callable type hint like `Awaitable[str]` are incommensurable. The type hierarchies they describe are completely unrelated.

## There's Probably a Point to All This?

Shockingly, there is. Because type hints only constitute a partial ordering under subtype compatibility, we can't apply the `@functools.total_ordering` decorator. Instead, we'll need to manually define each of Python's six rich comparison operators ourselves:

* `__eq__()`.
* `__ne__()`.
* `__ge__()`.
* `__gt__()`.
* `__le__()`.
* `__lt__()`.

This should still *mostly* be trivial. Three of those can be readily defined in terms of the other three: e.g.,

```python
def __ne__(self, other: TypeHint) -> bool:
    return not self.__eq__(other)

def __gt__(self, other: TypeHint) -> bool:
    return self.__ge__(other) and self.__ne__(other)

def __lt__(self, other: TypeHint) -> bool:
    return self.__le__(other) and self.__ne__(other)
```

That only leaves `__eq__()` and `__ge__()`, really.

*Annnnnyway...* Your work is brilliance and these are conundrums for another evening. For tonight we play video games until dawn and dream bleary-eyed of a creepy AI that will someday code all this for us and submit PRs with our usernames on them. :robot:

tlambert03:
> Okay. So that means type hints violate totality. If type hints satisfied totality, then either TypeHint(list[int]) <= TypeHint(Awaitable[str]) or TypeHint(Awaitable[str]) <= TypeHint(list[int]) would be True. But both conditions are False, because a container type hint like list[int] and a callable type hint like Awaitable[str] are incommensurable. The type hierarchies they describe are completely unrelated.

Yeah, I've been having this thought as I work on it.  Types are trees or graphs, not lines.  So, strictly speaking, will probably not make too much sense in _most_ cases.  I don't think it makes a huge difference though.  Ultimately, we're just using the `__le__` to mean "is this thing a subclass of other thing", just with the cute bonus of getting to use the `<=` operator.  But couldn't we just dispense with the `total_ordering` decorator altogether, and directly use an `is_subtype(other)` method... (and if we want to have a nice operator, we can still use `__le__` and `__ge__` to match python `set`'s `issubset` and `issuperset` behavior?

leycec:
**O frabjous day!** Your reign of excellence continues, I see. Now, let's see...

>  But couldn't we just... directly use an `is_subtype(other)` method... (and if we want to have a nice operator, we can still use `__le__` and `__ge__` to match python set's `issubset` and `issuperset` behavior?

Indeed, we can and we should do all those things! And I see that you have now done them. It's like you have a crystal ball into my brain, which is honestly kinda unsettling. :crystal_ball: 

Relatedly, as a user convenience, we might also consider publicly exposing this convenience tester you've helpfully privatized at the top of `beartype.math._mathcls`:

```python
def _is_subtype(subt: object, supert: object) -> bool:
    return TypeHint(subt) <= TypeHint(supert)
```

The `subt` and `supert` parameter names are a *bit* awkward for a public function. So, how about...

```python
def is_subtype(subtype: object, supertype: object) -> bool:
    '''
    Insert docstring for great glory here!
    '''

    return TypeHint(subtype) <= TypeHint(supertype)
```

> I removed total ordering in favor of directly making set-like methods (but kept the ordered operators).

**...excellent.** Your prowess as an API designer precedes you. <sup>*That's good.*</sup>

>  I wonder if we might want to soften some of the comment language about total ordering to use more set-like language...

**Indeedy!** Speaking of, would you mind globally replacing the substring `"total"` with `"partial"` everywhere in this PR? In theory, that should make all the wrong things right again.

Technically, the set of all possible type hints *is* a partially ordered countably infinite set. Pragmatically, Python doesn't do the whole "countably infinite set" thing out-of-the-box. It *does* do the whole "partially ordered" thing out-of-the-box. So, in the interest of clarity, I'd kinda prefer the latter if you don't mind too terribly much.

Also, there are some *really* interesting well-known algorithms and data structures one can apply to partial orders. Topological sorts and network clustering comes to mind. What could be funner? I mean, really. By highlighting that `beartype.math` defines a partial order over type hints, we're *really* highlighting that `beartype.math` enables type hints to be algorithmically reasoned with.

Indeed, and I'm not saying we should actually do this, but... We *could* augment `TypeHint` with `__add__()` and `__sub__()` dunder methods backed by `typing.Union[...]` type hints, in which case `beartype.math` would then be defining a partially ordered group `(G, +)` over type hints. Woah, bro! `beartype.math` would then expose the ponderous weight of [algebraic group theory](https://en.wikipedia.org/wiki/Group_theory) to the formerly  prosaic world of type hints.

Pinch me. I must be dreaming.

## Comments Continue

**Okay.** Let's get pragmatic, yo! Things we might want to do – possibly even in this PR:

* Define a trivial `TypeHint.__hash__()` dunder method in terms of the underlying type hint, which is (...more or less) mandatory whenever you define the `__eq__()` dunder method. The two go hand-in-hand: e.g.,

```python
    def __hash__(self) -> int:
        return hash(self._hint)
```

* Define a less trivial `TypeHint.__eq__()` dunder method. The current approach is super-clever, because it just defers to various `__eq__()` implementations in the `typing` module and C layer. Sadly, that's also the problem:

```python
>>> import typing as t

# This is good.
>>> t.List[str] == t.List[str]

# This is less good.
>>> list[str] == t.List[str]
```

An alternative approach would be for our `__eq__()` implementation to leverage (...*wait for it*) @beartype's `get_hint_pep_sign_or_none()` to equate semantically identical PEP 484- and 585-compliant type hints. Say, noice like:

```python
class TypeHint(object):
    ...

    def __eq__(self, other: object) -> Union[bool, NotImplemented]:

        # If that object is *NOT* an instance of the same class, defer to the
        # __eq__() method defined by the class of that object instead.
        if not isinstance(other, TypeHint):
            return NotImplemented
        # Else, that object is an instance of the same class.

        #FIXME: Maybe store as "self._hint_sign" to avoid lookups everywhere?
        # Sign uniquely identifying this and that hint if any *OR* "None"
        # (i.e., if either of these hints are *NOT* actually type hints).
        self_sign  = get_hint_pep_sign_or_none(self._hint)
        other_sign = get_hint_pep_sign_or_none(other._hint)

        # If either...
        if (
            # These hints have differing signs, these hints are unequal *OR*...
            self_sign is not other_sign or
            # These hints have a differing number of child type hints...
            len(self._hints_child_ordered) != len(other._hints_child_ordered)
        ):
            # Then these hints are unequal.
            return False
        # Else, these hints share the same sign and number of child type hints.

        # Return true only if all child type hints of these hints are equal.
        return all(
            self._hints_child_ordered[hint_child_index] ==
            other._hints_child_ordered[hint_child_index]
            for hint_child_index in range(self._hints_child_ordered)
        )


class _TypeHintClass(object):
    ...

    def __eq__(self, other: object) -> Union[bool, NotImplemented]:

        # If that object is *NOT* a partially ordered type hint, defer to the
        # __eq__() method defined by the class of that object instead.
        if not isinstance(other, TypeHint):
            return NotImplemented
        # Else, that object is a partially ordered type hint.
        #
        # If that hint is *NOT* a class type hint, these hints are unequal.
        elif not isinstance(other, _TypeHintClass):
            return False
        # Else, that hint is a class type hint.

        # Return true only if these hints describe the same class.
        return self._hint is other._hint
```

* Decorate `TypeHint.__new__()` by `@callable_cached`, which *should* then ensure that the `TypeHint()` factory only efficiently produces singletons: e.g.,

```python
from beartype._util.cache.utilcachecall import callable_cached

class TypeHint(object):
    ...

    @callable_cached
    def __new__(cls, hint: object) -> "TypeHint": ...
```

Given that, this constraint should now be true for any arbitrary type hint `hint`:

```python
>>> TypeHint(hint) is TypeHint(true)
True
```

* There's a slight complication with respect to our rich comparison operators deferring to the `is_subtype()` and `is_supertype()` methods. Don't get me wrong! Those methods are great. The PEP standard for rich comparison operators is a bit... wonky, however. Basically, those operators should return `NotImplemented` rather than raise an exception when the passed object is *not* an instance of `TypeHint`. Returning `NotImplemented` is the polite thing to do, because it allows the other object to have a say in the matter. Those methods are fine as is, thankfully. Sadly, we'll have to do this the manual way in those operators: e.g.,

```python
    def __le__(self, other: object) -> bool:
        """Return true if self is a subtype of other."""
        if not isinstance(other, TypeHint):
            return NotImplemented

        return self.is_subtype(other)
```

*le sigh.*

* `repr()` strings should, ideally, be evaluable as Python expressions. Doing so makes things a little more consistent across the Python ecosystem, which is nice. Happily, that's trivial here: e.g.,

```python
    def __repr__(self) -> str:
        return f"TypeHint({repr(self._hint)})"
```

And now for the big guns...

## `typing.Annotated` or Bust

In #137, I suggested many things. Let's not do those things. Instead, let's do this:

```python
from beartype._util.hint.pep.proposal.utilpep593 import (
    get_hint_pep593_metadata,
    get_hint_pep593_metahint,
)

class _TypeHintAnnotated(TypeHint):

    #FIXME: Define __eq__() in a similar way, probably? Joy is type hints.
    def _is_le_branch(self, branch: TypeHint) -> bool:

        # If that hint is *NOT* a "typing.Annotated[...]" type hint, this hint
        # *CANNOT* be a subtype of that hint. In this case, return false.
        if not isinstance(branch, _TypeHintAnnotated):
            return False
        # Else, that hint is a "typing.Annotated[...]" type hint.

        # Child type hints annotated by these parent "typing.Annotated[...]"
        # type hints (i.e., the first arguments subscripting these hints).
        self_metahint = get_hint_pep593_metahint(self._hint)
        other_metahint = get_hint_pep593_metahint(branch._hint)

        # Tuples of zero or more arbitrary caller-defined objects annotating by
        # these parent "typing.Annotated[...]" type hints (i.e., all remaining
        # arguments subscripting these hints).
        self_metadata = get_hint_pep593_metadata(self._hint)
        other_metadata = get_hint_pep593_metadata(branch._hint)

        # If either...
        if (
            # The child type hint annotated by this parent hint does not subtype
            # the child type hint annotated by that parent hint *OR*...
            TypeHint(self_metahint) > TypeHint(other_metahint) or
            # These hints are annotated by a differing number of objects...
            len(self_metadata) != len(self_metadata)
        ):
            # This hint *CANNOT* be a subtype of that hint. Return false.
            return False

        # Attempt to...
        #
        # Note that the following iteration performs equality comparisons on
        # arbitrary caller-defined objects. Since these comparisons may raise
        # arbitrary caller-defined exceptions, we silently squelch any such
        # exceptions that arise by returning false below instead.
        try:
            # Return true only if these hints are annotated by equivalent
            # objects. We avoid testing for a subtype relation here (e.g., with
            # the "<=" operator), as arbitrary caller-defined objects are *MUCH*
            # more likely to define a relevant equality comparison than a
            # relevant less-than-or-equal-to comparison.
            return all(
                self_metadata[metadata_index] == other_metadata[metadata_index]
                for metadata_index in range(self_metadata)
            )
        except:
            pass

        # Else, one or more objects annotating these hints are incomparable. So,
        # this hint *CANNOT* be a subtype of that hint. Return false.
        return False
```

In theory, that should let us compare beartype validators against one another. Will it? You decide.

**Okay.** So, let's talk type hints. Generally speaking, there are two kinds:

* **Unstructured type hints,** subscripted (indexed) by zero or more child type hints. Order is important, but there's *no* internal structure or implied semantics in that order. Most type hints are unstructured – like, 90% or more of all type hints, easily. Examples include unions and *most* type hints with an origin. Given an unstructured type hint, calling @beartype's low-level `get_hint_pep_args()` getter returns a semantically meaningful result.
* **Structured type hints,** subscripted (indexed) by two or more child type hints. Order is important. However, there's now an internal structure with implied semantics in that order. Examples include `typing.Annotated[...]`, `typing.Callable[...]`, `typing.(Async|)Generator[...]`, and probably a few more? You probably see where this is going. Given a structured type hint, calling @beartype's low-level `get_hint_pep_args()` getter mostly **does not return a semantically meaningful result.**

This isn't a bad thing, though. `typing.get_args()` tries hard to pretend that structured type hints are unstructured, which then causes problems elsewhere. `get_hint_pep_args()` doesn't bother pretending. Structured type hints are structured. It's best to preserve that structure, right? There are useful semantics in that structure.

Instead, @beartype provides a suite of higher-level getters retrieving that structure. These getters are specific to each category of type hint. Above, we saw the `get_hint_pep593_metadata()` and `get_hint_pep593_metahint()` getters specific to PEP 593-style `typing.Annotated[...]` type hints.

B-b-but what about...

## `typing.Callable` or Bust

**Busted.** We haven't implemented getters for `Callable[...]` type hints yet. Because of that, let's defer handling `typing.Callable` and `callable.abc.Callable` for a subsequent PR. Would that be okay with you?

It'll take me a day or two to doctor those getters up. There are at least three PEPs and more than a few edge cases across varying Python versions involved. I'd like to nail those unit tests properly. Once complete, these getters will resemble:

* A new `beartype._util.hint.pep.proposal.pep484585.utilpep484585callable.get_hint_pep484585_callable_args()` returning either:
   * A tuple of the zero or more parameter type hints subscripting (indexing) the passed `Callable[...]` type hint.
   * `Ellipsis`, if that hint was of the form `Callable[..., {return_hint}]`.
   * `typing.ParamSpec[...]`, if that hint was of the form `Callable[typing.ParamSpec[...], {return_hint}]`.
* A new `beartype._util.hint.pep.proposal.pep484585.utilpep484585callable.get_hint_pep484585_callable_return()` returning the return type hint subscripting (indexing) the passed `Callable[...]` type hint.

For similar reasons, I hope you don't mind if I close issue #137 out on you. Surprisingly, the current behaviour of `get_hint_pep_args()` is (...*mostly*) working as intended. Let me know if you hit any other deep snags and we'll bail you out with our pitifully small and leaky bailing can.

Phew!

## And... We're Almost There

So, you're in total and complete control over this PR. You control the horizontal and the vertical. Do as little or as much as you'd like. When you're ready, just let me know and **I'll immediately merge this.**

This is stellar output, @tlambert03. Colour me incredibly impressed. Just like me, this decorative emoji with mouth agape is both bald and in stunned awe. :open_mouth:

tlambert03:
howdie @leycec!  Just checking in.  I've been a bit busy with other stuff (and took the opportunity to let you do some stuff with `Callable`, but still very excited about this feature!  If callable stuff is ready-ish, or if I should just ignore callable for now and prepare this for final review, let me know and I'll get back on it.  (I think the main thing we needed from callable was the ability to detect `Callable[[], Any]` vs `Callable[..., Any]`?)

leycec:
@tlambert03, my favourite Harvard homie! You... you still live! Well, thank the Goddess. Kinda worried you had a heat stroke out on the golfing green there. :golf: :golfing: 

I have, of course, done everything. Okay, okay. It's all lies. I just implemented a single mostly untested `Callable` getter on your behalf:

```
>>> from beartype._util.hint.pep.proposal.pep484585.utilpep484585callable import (
...        get_hint_pep484585_callable_args)
>>> from beartype.typing import Any, Callable, ParamSpec
>>> get_hint_pep484585_callable_args(Callable[[], Any])}
()
>>> get_hint_pep484585_callable_args(Callable[[()], Any])}
()
>>> get_hint_pep484585_callable_args(Callable[[int], Any])}
(<class 'int'>,)
>>> get_hint_pep484585_callable_args(Callable[[int, bool], Any])}
(<class 'int'>, <class 'bool'>)
>>> get_hint_pep484585_callable_args(Callable[..., Any])}
Ellipsis
>>> get_hint_pep484585_callable_args(Callable[ParamSpec("P"), Any])}
~P
```

I'm convinced. Let's pretend that callable stuff is ready-ish... *ish.* I defer to your magnanimous wisdom as Master of the Type Hint Math. Just lemme know when you're exhausted and on the verge of chucking your keyboard through the bathroom window. At that crisis point, I'll happily merge everything.

And thanks so much for revisiting this profound work of human genius. May your Summer be filled with pool noodles, endless intoxication, and the sweet scent of suntan lotion mingled with reusable chlorine pucks. :sun_with_face: :tropical_drink:

leycec:
:heart_on_fire:

tlambert03:
OK @leycec, this is ready for final review!  Notable exceptions at the moment include TypeVar, ParamSpec, and Generics.

They raise informative errors and I think this PR is useful enough as is to tackle those in the future.  Test coverage is at ~100%.  Please pay the most attention to the `__new__` and `__init__` methods... where your bear-expertise would be greatly appreciated.

Looking forward to using this!

leycec:
**Wondrous day!** I was *just* about to push out `beartype 0.11.0` tonight, but let's merge your fabulous PR instead. That way, we can say we actually did something in `beartype 0.11.0`. By "we," I of course mean you. :smiling_face_with_tear:

leycec:
**The post-merge review has landed.** As the years furiously grind on, I've come to devalue the meticulous sort of line-by-line peer code review that I used to value. That sort of review exhausts everyone, depletes precious volunteer enthusiasm, and always makes me feel like a Big Boss Bad Dude.

Instead, I'll just splatter this comment with a few ad-hoc observations:

* Generally speaking, **your work is stellar.** Seriously. I'm so grateful.
* Test submodules like `beartype_test.test_is_subtype` should *usually* shy away from performing global imports at module scope. Instead, all test-specific imports should *usually* be isolated to the body of each test requiring those imports. Why? Because test submodules performing global imports at module scope **will** grind the entire `pytest` process to a halt if even a single one of those imports fails. Obvious exceptions to this heuristic include imports from Python's standard library like `collections.abc` and `typing`. Those are mostly *always* safe. Imports from the main @beartype codebase, however, are mostly *never* safe. In this case:

```python
# These global imports from @beartype in "test_is_subtype"...
from beartype.math._mathcls import (
    _HINT_SIGNS_ORIGIN_ISINSTANCEABLE_ARGS_1,
    _HINT_SIGNS_ORIGIN_ISINSTANCEABLE_ARGS_2,
    _HINT_SIGNS_ORIGIN_ISINSTANCEABLE_ARGS_3,
    TypeHint,
    is_subtype,
)
from beartype.roar import BeartypeMathException

# ...should instead be isolated to each test that needs them.
# For example:
def test_invalid_subtype_comparison():
    # Defer test-specific imports.
    from beartype.math._mathcls import TypeHint
    from beartype.roar import BeartypeMathException

    hint = TypeHint(t.Callable[[], list])
    with pytest.raises(
        BeartypeMathException, match="not a 'beartype.math.TypeHint' instance"
    ):
        hint.is_subtype(int)
```

**No biggie,** obviously. I'll happily repair that a bit. Since I'm obsessively fastidious about tests, this kind of laborious tedium is in my wheelhouse. *Next!*

* **Tests are failing under at least Python 3.7 and 3.8.** Unfortunately, they're failing due to aforementioned global imports at module scope in the `test_is_subtype` submodule – which means tests are *really* failing. Fortunately, it looks like the culprit is pretty simple: `beartype.typing.Protocol` is only defined under Python ≥ 3.8, which explains the Python 3.7 test failure. The Python 3.8 test failure is presumably similar? In any case, I'll just brutally comment out the entire `test_is_subtype` submodule for now. It hurts! Let's repair that at some time later this week, if we can. For now, *laziness prevails*. :weary: 
* For the aforementioned reason, consider locally running tests via `tox`: e.g.,

```bash
$ cd ~/py/beartype
$ ./tox
...
___________________________________________________________ summary ___________________________________________________________
  py37-coverage: commands succeeded
  py38-coverage: commands succeeded
  py39-coverage: commands succeeded
  py310-coverage: commands succeeded
SKIPPED:  py311-coverage: InterpreterNotFound: python3.11
SKIPPED:  pypy37-coverage: InterpreterNotFound: pypy3.7
SKIPPED:  pypy38-coverage: InterpreterNotFound: pypy3.8
SKIPPED:  pypy39-coverage: InterpreterNotFound: pypy3.9
  congratulations :)
```

As the summary suggests, `tox` exercises your changes against *all* installed Python interpreters. It's positively magical! The downside, of course, is that you'll have to actually install all of those Python interpreters on your local machine. The [third-party open-source `pyenv` package](https://github.com/pyenv/pyenv#install-additional-python-versions) is *usually* how most non-crazy people make that happen. I, of course, use Gentoo Linux instead. So, pretty crazy. Don't be like me. Just use `pyenv` to install Python 3.7, 3.8, and 3.9: e.g.,

```bash
$ curl https://pyenv.run | bash  # <-- *GULP* shell code injection *GULP*
$ pyenv install 3.7
$ pyenv install 3.8
$ pyenv install 3.9
```

I fully support doing dangerous things at the terminal.

* The `SUBTYPE_CASES = [...]` data structure is **h0tness.** Seriously. I totally get why that's global; it's really gotta be to be useful. Excellent test automation deserves excellent emoji applause. :clap: :clap: :clap:
* The `test_type_hint_singleton()` and `test_arg_nparams()` tests are also **h0tness,** clearly.
* The `test_callable_takes_args()` test concerns me a bit. I see we're publicizing public `TypeHint.takes_no_args` and `TypeHint.takes_any_args` instance variables for callable type hints. I mean, I *get* it. Definitely. But it also kinda seems like we're edging uncomfortably close to reproducing the [PEP 362-compliant function signature API](https://peps.python.org/pep-0362) (i.e., [the `inspect.Signature` class](https://docs.python.org/3/library/inspect.html#introspecting-callables-with-the-signature-object)) here. There's probably a good reason for that – but maybe we just want to go "whole hog" and publicize a public `TypeHint.signature = inspect.Signature(...)` instance variable instead that reuses that standard API? Food for thought tastes good.
* The `test_callable_takes_args()` test also concerns me a bit. I see we're publicizing a public `TypeHint.is_empty_tuple` instance variable for fixed-length tuple type hints. Thankfully, I see that we're also defining:
  * A private `TypeHint._args` instance variable. Given that, a general-purpose alternative might be to:
    * Make that public by renaming `TypeHint._args` to `TypeHint.args`.
    * Replace `TypeHint.is_empty_tuple` with `len(TypeHint.args) == 0` or even just `not TypeHint.args`. Both of those are fairly readable and more widely applicable to similar sequence type hints. *Annnnnyway!* Just a passing thought in the wind of the night.
  * Alternately, it would probably make even *more* sense to define a `TypeHint.__len__()` dunder method. Since you've already *amazingly* defined the `TypeHint.__iter__()` dunder method, we'd then satisfy both the `collections.abc.Iterable` and `collections.abc.Sized` protocols – which would almost (but not quite) make `TypeHint` objects full-blown collections satisfying `collections.abc.Collection`. In any case, defining `TypeHint.__len__()` would then let us test for empty sequence type hints with even *more* Pythonic `len(TypeHint) == 0` and `not TypeHint` tests. That seems valuable at some distant date that I am too exhausted to even contemplate.
* **Type hint enumeration is...** wild, crazy, and outrageous! I didn't expect this, but now I can't live without it:

```python
    >>> list(TypeHint(Callable[Union[int, str], Sequence[Any]]))
    [TypeHint(typing.Union[int, str]), TypeHint(typing.Sequence[typing.Any])]
```
* This fragment in `TypeHint.__new__()` seems a little magical:

```python
            if isinstance(hint, type) or getattr(hint, "__module__", "") == "typing":
                TypeHintSubclass = _TypeHintClass
```

The first condition `isinstance(hint, type)` makes absolute sense, but the second condition... That's kinda intense. Should we *really* be conflating `typing` attributes that aren't types with objects that are types?

Oh, and it's germane to note here that @beartype actually has a tester for testing whether type hints are `typing` attributes: `beartype._util.hint.pep.utilpeptest.is_hint_pep_typing()`. That test is shockingly non-trivial, because @beartype also needs to transparently support the third-party `typing_extensions` module (as much as reasonably can be, anyway). So, we might rewrite the above to resemble:

```python
            from beartype._util.hint.pep.utilpeptest import is_hint_pep_typing
            if isinstance(hint, type) or is_hint_pep_typing(hint):  # <-- ...still unsure about this
                TypeHintSubclass = _TypeHintClass
```

And... I'm exhausted. I hardly even touched the bulk of your brilliance. But let's let sleeping dogs lie so that I can sleep, too. PR #136, I merge you!

leycec:
:boom: :boom: :boom:

leycec:
**Yay!** Our test suite once again passes – except for `pyright`, of course, which yells nonsensically at the top of its perfidious lungs about four non-issues:

```
No configuration file found.
No pyproject.toml file found.
Assuming Python platform Linux
Searching for source files
Found 180 source files
pyright 1.1.254
/home/leycec/py/beartype/beartype/math/_mathcls.py
  /home/leycec/py/beartype/beartype/math/_mathcls.py:25:33 - error: "ParamSpec" is unknown import symbol (reportGeneralTypeIssues)
  /home/leycec/py/beartype/beartype/math/_mathcls.py:476:72 - error: Argument of type "Unknown | None" cannot be assigned to parameter "__class_or_tuple" of type "type | tuple[type | tuple[Any, ...], ...]" in function "isinstance"
    Type "Unknown | None" cannot be assigned to type "type | tuple[type | tuple[Any, ...], ...]"
      Type "None" cannot be assigned to type "type | tuple[type | tuple[Any, ...], ...]"
        Type "None" cannot be assigned to type "type"
        Type "None" cannot be assigned to type "tuple[type | tuple[Any, ...], ...]" (reportGeneralTypeIssues)
  /home/leycec/py/beartype/beartype/math/_mathcls.py:476:72 - error: Second argument to "isinstance" must be a class or tuple of classes
    TypeVar or generic type with type arguments not allowed (reportGeneralTypeIssues)
  /home/leycec/py/beartype/beartype/math/_mathcls.py:579:44 - error: Index 0 is out of range for type tuple[()] (reportGeneralTypeIssues)
4 errors, 0 warnings, 0 informations 
Completed in 7.873sec
```

Silence, `pyright`! *Or taste bloody steel.* :crossed_swords:

tlambert03:
hooray!  Thanks for all the comments and tips... Even though you generously merged anyway, I'll follow up with some improvements based on these suggestions

leycec:
You're most welcome – and thanks *yet* again for this Big Bump to @beartype. :hugs: 

Please punch my avatar if I accidentally stomp on your parallel efforts, but I'll be massaging the `beartype.math` subpackage myself a *heap* tonight. Priorities include:

* **Renaming `beartype.math`.** N-n-now... hear me out here. I came up with a ludicrous acronym and we're going to learn to live with it: the **D**ecidely **O**bject-**O**rientedly **R**ecursive API. Or, `beartype.door` for short. Open the door to a whole new type-hinting world. :door: 
* **Moving `beartype_test.test_is_subtype` somewhere deeper and more secretive in the creaky bowels of our test suite. More on that as I actually do something.
* **Resolving all pending false positives from `pyright`.** The less everyone says about `pyright`, the better.
* **Splitting `beartype.door._mathcls` into subsidiary submodules for maintainability – but mostly just to assuage my crippling obsessive-compulsive disorder.

Oh, and I just stumbled over [@napari](https://github.com/napari/napari) on my way to GitHub work today. I am bedazzled. If @napari isn't the most impressive biomedical-oriented 3D Python viewer *evar*, I will livestream our Maine Coon eating my Tilley Hat sprinkled with catnip. Permissive licensing is only the cherry on top of your world-class crêpe. Educated prediction: @napari dominates client-side 3D visualization in 2022. Just... **wowza.** :open_mouth:

leycec:
**Catch!** :basketball: 

If you'd like, the ball's in your court. Tests still fail under Python 3.7 and 3.8, but we're considerably closer. When you get a free minute, <sup>*...so never is what I'm saying*</sup> would you mind taking a look at [the failing `test_is_subhint()` test under Python 3.8](https://github.com/beartype/beartype/blob/d74a4f05819cbd6d9efcc89d92ecaba33e375e53/beartype_test/a00_unit/a60_api/door/test_door.py#L235)?

Something smells disgustingly fishy there. It might be me. But it might be the code.

leycec:
I can confirm that the `test_is_subhint()` test passes under Python ≥ 3.9. :partying_face: 

Until we can decipher what exactly went wrong under older Python versions, our test suite now conditionally ignores that test under Python < 3.9 – ensuring tests continue to run under at least Python ≥ 3.9.

Let's see what insights a debauched Friday evening brings.

leycec:
**Test party continues.** `test_typehint_new()` now unconditionally passes everywhere and `test_typehint_equals()` now passes under Python ≥ 3.9. Python 3.7 and 3.8 remain hostile to our interests. *What you gonna do?*

<sup>Gods, but I hope the answer is: "Go to sleep, you."</sup>

leycec:
**API banging continues.** I'd just like to reiterate, @tlambert03, that your work here was sufficiently phenomenal that it convinced someone they were wrong. This someone is me. How often does *that* happen in real life? Less often than we would like.

So, what was I wrong about? I was wrong to neglect the object-oriented approach. Instead, I implemented @beartype 100% imperatively. Sure, there are a handful of dataclasses percolating about – but the code generation API at the heart of @beartype is almost purely procedural. That's good, in the sense that that's as wickedly performant as pure-Python can be; but that's also bad, in the sense that that's increasingly unusable, unreadable, and unmaintainable. I'll take a maintainable decorator over a performant decorator.

## The Future Beckons with a Creepy White Glove

**Enter your API,** stage left. I'm dumbfounded at how blatantly superior this API is to everything else I invented. It's Pythonic, it's usable, it's readable, it's maintainable, and it's on the cusp of *just working*.

What I'm trying to say here is that @beartype will eventually replace *everything* it internally does with this API instead. To do so, I'll be slowly building out support for efficient `O(1)` runtime type-checking directly in the `beartype.door.TypeHint` class and subclasses thereof.

This will probably take a year or two. But I'm committed to making this happen, because this *has* to happen. Sanity demands a sane API. Thus spake the Bear. :bear:

tlambert03:
wowie!  that's big news.  I'm glad to hear you're finding some inspiration leaking out of this PR into other areas.

I agree, this pattern did end up feeling maintainable and relatively easy to reason about (breaking up the recursion into little digestible bits).  ... and I guess it's not a huge leap to just kinda add an `is_instance` concept to a `TypeHint` for runtime checking?  That would indeed be a nice API.

Performance is nothing to shrug off though, so I look forward to seeing what your uncompromising nature comes up with as an ideal marriage of the two :)

leycec:
**Sadness ensues,** because Python 3.7 and 3.8 are still failing. I'm currently re-enabling each test in the `beartype_test.a00_unit.a60_api.door.door_test` submodule one-by-one, repairing those tests that can be trivially resolved to support Python 3.7 and 3.8 and conditionally skipping the rest with `@skip_if_python_version_less_than('3.9.0')`.

If you find a spare weekend between this and your other open-source commitments, <sup>[napari](https://napari.org/) :open_mouth: [napari](https://napari.org/)</sup> it'd be super-swell if you could investigate the `test_is_subhint()` and `test_typehint_equals()` tests. We're skipping both with `@skip_if_python_version_less_than('3.9.0')` at the moment, as I couldn't quite grok what was happening there.

But at least `pyright` has been manhandled into submission! We give thanks to the eternal QA diety, Pacha Lubre, for his oversight in this matter.

tlambert03:
will take a look!
by the way, can you give me a quick rundown on the beartype policy with `typing_extensions`? (apologies, I'm sure this must be described somewhere, but I couldn't find it after a few searchers).  It seems to generally be the case that you avoid supporting/testing backported types on earlier python versions, yes?  So, if I run into a `typing` object that wasn't introduced until some version, I shouldn't try to test the typing_extensions backport?

leycec:
*...heh.* Our policy is simple. Exactly as you surmise, we gratefully appreciate the existence of `typing_extensions` while yet loathing the runtime implementation of `typing_extensions`. @beartype only adds explicit support for specific `typing_extensions` type hints when someone suddenly opens up an issue heckling us about our lack of said support.

Currently, this means:

* `typing_extensions.Annotated`, which we *must* support to enable the sad subset of our userbase still locked into Python 3.7 and 3.8 to transparently use @beartype validators.
* `typing_extensions.Literal`, because the [friendly `graspologic` guys at Microsoft](https://github.com/microsoft/graspologic) *really* wanted us to do that. We gritted teeth and quietly did that. We do this for you, Microsoft! You're welcome.

All other `typing_extensions` attributes are mostly unsupported. They might work; they might not work. Usually more the latter than the former. Praise be to RNG Jesus!

scary jeebus

**Let us now explain why.** Historically, PEP authors have tended to their trial-run unfinished beta implementations of upcoming type hint factories (that are currently under peer review by the `python-typing` mailing list community) within `typing_extensions` itself. Once a PEP is accepted, those unfinished beta implementations *never* get updated within `typing_extensions`.

We see where this is going. It's pure anarchy in the streets! Most of `typing_extensions` either doesn't work at runtime *or* sorta works but in a completely non-standard way that's broken af from the perspective of runtime standards compliance. In other words...

> So, if I run into a `typing` object that wasn't introduced until some version, I shouldn't try to test the `typing_extensions` backport?

**Please God(s),** don't test the `typing_extensions` backport. :sweat_smile:

tlambert03:
ok, the issue with python 3.7 and 3.8 appears is somewhat related to a difference between from `typing.get_args` and `utilpepget.get_hint_pep_args`... but it's also somewhat related to our current lack of implementation for `TypeVar`.

It appears that back in 37 and 38, the default argument to a bare, unparametrized generic like `typing.List` was [an instance of `TypeVar`](https://github.com/python/cpython/blob/3.7/Lib/typing.py#L1245).  In py39 and later, they did away with passing default params to _GenericAlias, and instead [started passing nparams instead](https://github.com/python/cpython/blob/3.9/Lib/typing.py#L1733).  They handled this internally with `typing.get_args`, such that if you use `typing.get_args(typing.List)` you get an empty tuple in all versions of python that had `get_args` (≥3.8).  However, beartype goes straight for the `__args__` attribute, and doesn't check if it's a special GenericAlias, so on py≤3.8 you'll get `get_hint_pep_args(List) == (~T,)` ... but on ≥ 3.9 you'll get `get_hint_pep_args(List) == ()`

### what it means for us

I can make all tests in `test_door.py` pass in both py37 and py38 just by editing 
https://github.com/beartype/beartype/blob/6582be541dd1b11dc27e8da3631fbbdc799764c4/beartype/_util/hint/pep/utilpepget.py#L90-L94

to read as:
```python
        if getattr(hint, '_special', False):
            return ()
        return getattr(hint, '__args__', ())
```

unfortunately, that then breaks some of your other tests `test_a00_utilpepget.py::test_get_hint_pep_args`, because it's expecting anything with ` hint_pep_meta.is_args` to have arguments.

I'm not really sure how you want to fix this one... here's the options I see:

1. in `doorcls`, we just act like TypeVars aren't even there for now... and treat `List[TypeVar("T")]` exactly the same as `List`
   - outside of this being a strictly temporary fix, I don't really like it, since it's ignoring the root cause.  Furthermore, once we do support `TypeVar`, we're gonna be right back up against this problem
2. deal with the root cause in get_hint_pep_args.  We kinda need to know if something is an unparametrized GenericAlias vs a parametrized one, and currently, in python 3.8 and less, we can't
    ```python
    In [4]: get_hint_pep_args(typing.List)
    Out[4]: (~T,)
    
    In [5]: get_hint_pep_args(typing.List[typing.TypeVar("T")])
    Out[5]: (~T,)
    ```
    note though, `typing` knows:
    
    ```python
    In [7]: get_args(typing.List)
    Out[7]: ()
    
    In [8]: get_args(typing.List[typing.TypeVar("T")])
    Out[8]: (~T,)
    ```
    I think I need you for this one though, since I'm not sure how you want to deal with the failing test and `hint_pep_meta.is_args`
3. just leave these tests failing until we fully implement support for `TypeVar`.  Since an unbound `TypeVar` is, I guess, the same as `Any`, we don't _really_ need to know whether we're looking at a parametrized vs unparametrized generic... but we do need to otherwise fully analyze the bounds of the `TypeVar`.  That's a lot of work for a little problem here... but ultimately, we should tackle type var. 

thoughts?

leycec:
**Breathtaking writeup!** You continue to both stun and impress. The answer, of course, is **2.** Right? You knew that was coming. @beartype is here to do things semantically right, even when that means doing things syntactically wrong. `get_hint_pep_args()` already ~~lies about~~ gently massages a few syntactic things for the greater good of semantic correctness. Let's keep doing that.

Moreover, we really *have* to keep doing that. I'd never actually noticed that our current approach prevented us from distinguishing an unsubscripted `List` from a parametrized `List[TypeVar('T')]`. But... apparently, it does. That's horrifying! Even our balding Maine Coon is appalled.

Thankfully, you even solved that the Right Way™ too. Let's temporarily break tests by implementing your proposed solution ala:

```
# In "utilpegget":

...
# Else, the active Python interpreter targets Python < 3.9. In this case,
# implement this getter to directly access the "__args__" dunder attribute.
else:
    def get_hint_pep_args(hint: object) -> tuple:

        # If this hint is unsubscripted (e.g., a bare "typing.List" reference),
        # return the empty tuple. Note this edge case is:
        # * Detected and handled by the standard "typing" implementation
        #    under Python 3.7 and 3.8 in the same exact way.
        # * Required to differentiate unsubscripted hints from parametrized
        #    hints (e.g., "List[TypeVar('T')]"). Why? Because the "typing"
        #    module nonsensically implements unsubscripted hints as
        #    parametrized hints under Python 3.7 and 3.8. The only safe
        #    means of disambiguating these two common cases is to
        #    violate privacy encapsulation by detecting this instance variable
        #    set to "False" for unsubscripted but *NOT* parametrized hints.
        if getattr(hint, '_special', False):
            return ()
        # Else, this hint is subscripted by one or more child type hints...
        # probably. But nothing is certain. We remain vigilant yet fearful.

        # Return the value of the "__args__" dunder attribute if this hint
        # defines this attribute *OR* the empty tuple otherwise.
        return getattr(hint, '__args__', ())
```

Does that sound reasonable? Honestly, I can't thank you enough – but I will anyway. :clap: :partying_face: :clap:

leycec:
**Callooh! Callay!** I'm *fairly* sure (but technically uncertain) that tests that fail after applying your wondrous fix can be trivially repaired by:

* In `beartype_test.a00_unit.a60_api.data_pep484`:

```python
    # Let's change this line...
    _IS_ARGS_HIDDEN = _IS_TYPEVARS_HIDDEN

    # ...to resemble this instead. Sanity, we restore thee.
    _IS_ARGS_HIDDEN = False
```

That's it. I think? Let's see how far the rabbit hole goes. :hole: :rabbit2:

PR #145, opened 2022-07-16
Fix doortest for py37 py38

tlambert03:
This fixes test_door for python 3.7 and 3.8 by modifying `get_hint_pep_args` to return the empty tuple if the hint is a special generic alias (i.e. has the `_special` attribute set to `True`)

see explanation for why `get_args` was involved here in https://github.com/beartype/beartype/pull/136#issuecomment-1186037251

### DIFF
--- a/beartype_test/a00_unit/a60_api/door/test_door.py
+++ b/beartype_test/a00_unit/a60_api/door/test_door.py
@@ -77,7 +77,6 @@ def hint_subhint_cases() -> 'Iterable[Tuple[object, object, bool]]':
         Sequence,
         Sized,
         Tuple,
-        TypedDict,
         NamedTuple,
         Union,
     )
@@ -91,11 +90,6 @@ def hint_subhint_cases() -> 'Iterable[Tuple[object, object, bool]]':
     class MuhNutherThing:
         def __len__(self) -> int:
             ...
-
-
-    class MuhDict(TypedDict):
-        thing_one: str
-        thing_two: int
 
 
     class MuhTuple(NamedTuple):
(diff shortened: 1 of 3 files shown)


beartype_test/a00_unit/a60_api/door/test_door.py
```diff
@@ -230,12 +226,12 @@ def muh_method(self):
             (MuhNutherThing, MuhThingP, False),
             (MuhThingP, MuhThing, False),
             # PEP 586-compliant type hints.
-            (Literal[1], int, True),
+            (Literal[7], int, True),
```
tlambert03, line 229: have to change this from Literal[1] to something else because (oddly enough) it was breaking a check on the exception message over in codemain because of type caching in the typing module.  Whichever of `Literal[True]` and `Literal[1]` gets defined first becomes the permanent repr for _either_ of them! :joy:

leycec: **ZOMG.** That's hilarious... and horrible! I'd just added this test case to the `test_door` submodule, too:

```python
        # Blame Guido.
        (bool, int, True),
```

Like, seriously. Law of Unintended Consequences strikes back.

`hash(1) == hash(True)` is such a festering pile of Bad Things™, because it all but guarantees hash collisions for anything mixing booleans and integers. How did Python 3.x never resolve that? But this...

> ...whichever of `Literal[True]` and `Literal[1]` gets defined first becomes the permanent repr for *either* of them! :joy:

Python: **AM I A JOKE TO YOU!?**

Guido: ...heh

leycec:
**Unadulterated brilliance.** You personify the good in mankind. I'm so incredibly grateful for your continued awesomeness. The @beartype Volunteer Resources (VR) division encourages you take the month off of everything except @napari and [`magicgui`](https://github.com/napari/magicgui) – which is also awesome, because of course it is. @tlambert03 did it.

We will praise you loudly in our changelog for the upcoming `beartype 0.11.0` push. Your name will ring from the GitHub commit heavens throughout time and space. Thanks for everything, Talley. This squash merge's for you. :hugs:

PR #148, opened 2022-07-18
feat: add NewType Support to TypeHint

tlambert03:
support NewType, simply by checking the `__supertype__` attribute.  See comment about a future "strict" mode (that would further assert NewType identity) in the PR.

### DIFF
--- a/beartype/door/_doorcls.py
+++ b/beartype/door/_doorcls.py
@@ -46,6 +46,9 @@ from beartype._util.hint.pep.utilpepget import (
     get_hint_pep_origin_or_none,
     get_hint_pep_sign_or_none,
 )
+from beartype._util.hint.pep.proposal.pep484.utilpep484newtype import (
+    get_hint_pep484_newtype_class,
+)
 from beartype._util.hint.utilhinttest import is_hint_ignorable
 from contextlib import suppress
 
@@ -193,7 +196,7 @@ class TypeHint(ABC):
         # Else, this hint is supported.
 
         # If a subscriptable type has no args, all we care about is the origin.
-        if not get_hint_pep_args(hint):
+        if not get_hint_pep_args(hint) and hint_sign not in {HintSignNewType}:
             TypeHintSubclass = _TypeHintClass
 
         # Return this subclass.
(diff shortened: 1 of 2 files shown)


leycec:
**Wondrous birth of a new miracle,** but @tlambert03 did it again. You have my undying gratitude for deferring to our `get_hint_pep484_newtype_class()` getter, too. How did you even find that buried treasure? You possess uncanny code discovery skills even I have never fathomed.

Let's commit this for the communal good of mankind. Party like it's Python 1.0.0 all over again! :pinata:

tlambert03:
> So, abstract syntax tree (AST) inspection probably.

or... I guess I what I was thinking is that you could have an option that more strictly asserts that some type hint _IS_ some NewType (rather than checking subclass).  

i.e. 
```python
S = NewType("S", str)
assert is_subhint(S, str)
assert is_subhint(S, S)
assert not is_subhint(str, S)  # currently this one is also True
```

leycec:
**Oh.** Right. I must be misremembering. As I dimly recall:

* Under Python < 3.10, `NewType` was implemented as a closure accepting the bizarrely named parameter `x` that just returned `x` as is. Yeah. Cray-cray.
* Under Python ≥ 3.10, `NewType` was refactored to be an actual type whose `__call__()` dunder method again accepts and returns `x` as is. Also cray-cray, but less so.

**Ah-ha.** I gets it, finally. Nevermind everything above. So, we maybe want to implement `NewType` internally by dynamically fabricating a new subclass of the `__supertype__` rather than just reducing `NewType` to that `__supertype__`. In theory, that should then ensure that that `NewType` is *not* a subhint of that `__supertype__`: e.g.,

```python
# Let's treat this...
NewType('S', str)

# ...as semantically equivalent to this.
class S(str): pass
```

Incidentally, you're amazing. So impressed by your uncanny cunning and relentless correctness. :sparkler:

tlambert03:
>  ...as semantically equivalent to this. Maybe?
class S(str): pass

ok... so the current behavior of `is_subhint(str, S)`  would be wrong then right?  I can get behind that.  I don't know why this one feels a little tricker to decide.  I guess if someone has gone through the trouble of hinting with `NewType` then their intention is probably to say "not just any old string" ... so you're right `type("S", (str,), {})` is probably the way to think about it.  Do you think this needs any sort of configurable?  Or do we just have an Opinion™️ and stick with it :)

> So impressed by your uncanny cunning and relentless correctness. 🎇

you flatter 😊  this is fun stuff to think about!

leycec:
>  guess if someone has gone through the trouble of hinting with NewType then their intention is probably to say "not just any old string" ...

Right? They went out of their way to semantically differentiate their type, so we should probably do the same. From that perspective, `NewClass = NewType(NewClassname, OldClass)` really just is statically parsable shorthand for the statically unparsable long form `NewClass = type(NewClassname, (OldClass,), {})`.

I cannot believe you know that trickery too, by the way. The mostly undocumented 3-parameter variant of the `type()` builtin is known to only a select cadre of elite GitHubbers. I feel less special now. :sob:

tlambert03:
> I cannot believe you know that trickery too, by the way. 

I like my magic as black as i can get it :joy:

PR #150, opened 2022-07-19
fix: treat newtype as subtype (not an alias) of it's supertype

tlambert03:
implements https://github.com/beartype/beartype/pull/148#issuecomment-1189450220

what do you think @leycec?  is it safe to do it this way using `type(...)?`  i figured since `TypeHints` are cached, it should work?

### DIFF
--- a/beartype/door/_doorcls.py
+++ b/beartype/door/_doorcls.py
@@ -1114,7 +1114,23 @@ class _TypeHintNewType(_TypeHintClass):
     # relatively fringe case.
     def __init__(self, hint: object) -> None:
         super().__init__(hint)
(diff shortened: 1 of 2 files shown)


leycec:
**Cleverness continues.** You even handled the edge case of an unsubclassable type, which is unthinkable madness *and yet it happens.* I agree that this looks totally safe in light of `TypeHint` caching, so let's pretend that everything will just work out if we wish upon a GitHub :star:.

Incidentally, you are the open-source barometer by which I now measure yourself. Your GitHub productivity exceeds my own by at least twenty orders of magnitude. I am humbled. It's a good feeling. Let's merge this before I cry into my beer.

PR #157, opened 2022-08-09
Fix CI not running

dosisod:
I noticed that the Github Actions have been failing for the last couple days:

image

When you click on the ones that are failing right way, they are saying:

```
Error when evaluating 'runs-on' for job 'tests'. .github/workflows/python_test.yml (Line: 171, Col: 14): Unexpected type of value '', expected type: OneOf.
```

Basically, the `include` line(s) are adding metadata to an existing matrix job (based on the key `pypy37-coverage`), but since the key is not specified in the `tox-env` list above, it creates a new, blank (!!) matrix job, hence why `${{ matrix.platform }}` is empty.

### DIFF
--- a/.github/workflows/python_test.yml
+++ b/.github/workflows/python_test.yml
@@ -155,10 +155,10 @@ jobs:
           #FIXME: Reduce to simply "3.11" after Python 3.11 goes live.
           # - tox-env: py311-coverage
           #   python-version: "3.11.0-alpha - 3.11.0"
-          - tox-env: pypy37-coverage
-            python-version: "pypy-3.7"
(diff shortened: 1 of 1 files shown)


leycec:
**Gah!** I failed to even notice that CI had been failing... *for a friggin' week.* I blame this summer heatwave that has me bathed in my own musky sweat. :sweat_drops:  

Thanks yet again for the tremendous outpouring of volunteerism, @dosisod. You truly are the open-source wizard amongst us. Merge this immediately, GitHub. :mage_man:

dosisod:
Of course! And thank you for making such an awesome library, I look forward to using it in some of my future projects!

Issue #161, opened 2022-08-27
[Feature Request] Visual separation between beartype text and the user input in error messages


justinchuby:
Currently beartype does not use symbols etc. to separate the user input from the rest of the error message. When the type violating value is long or its type atypical, it becomes harder for a user to read the error message.

Proposal:

1. Use a symbol to quote the values. The backtick `` ` `` symbol can be a good candidate
2. Use colors
3. Do both!
    <img width="970" alt="image" src="https://user-images.githubusercontent.com/11205048/187042220-037d151d-80ac-4ed1-98f3-6f34ffeb9532.png">

    A more complicated example showing the user input is clearly distinguishable:

    ![image](https://user-images.githubusercontent.com/11205048/187043486-3c5949e1-c2b9-4482-a5f6-5c969122048b.png)

> (1) I am hoping to color the correctly matched type green (`*List* index 0 ...`). (2) I wasn't able to find the code that outputs the first line `Not <class ...> or ...`

## Test script

```python
from beartype import beartype

@beartype
def quote_wiggum(lines: list[str] | str) -> None:
    print('“{}”\n\t— Police Chief Wiggum'.format("\n ".join(lines)))
    return lines

def print_exception(func):
    try:
        func()
    except Exception as e:
        print(e)

print_exception(lambda: quote_wiggum(['Hello', 'World']))
print("------")
print_exception(lambda: quote_wiggum([b"Oh, my God! A horrible plane crash!", b"Hey, everybody! Get a load of this flaming wreckage!",]))
```

yields

```
@beartyped __main__.quote_wiggum() return `['Hello', 'World']` violates type hint `None`, as list `['Hello', 'World']` not instance of <class "builtins.NoneType">.
------
@beartyped __main__.quote_wiggum() parameter `lines=[b'Oh, my God! A horrible plane crash!', b'Hey, everybody! Get a load of this flaming wrecka...'` violates type hint `list[str] | str`, as list `[b'Oh, my God! A horrible plane crash!', b'Hey, everybody! Get a load of this flaming wrecka...'`:
* Not str.
* List index 0 item bytes `b'Oh, my God! A horrible plane crash!'` not instance of str.
```

langfield:
[...]

PR #162, opened 2022-08-27
Colorize error messages

justinchuby:
Colorize error messages.

image

#161


leycec:
**Phenomenal!** I'm so tremendously apologetic for my recent radio silence. Open-source volunteer time is scarce. I know, right? With what little free time I have left, I've been cartwheeling like an ultra-marathon runner on the final mile towards the finish line that is `beartype 0.11.0`.

The good news is that I finished everything for `beartype 0.11.0` five minutes ago. The bad news is that I neglected this amazing PR and your companion issue at #161. But my stupidity ends now.

**Congratulations.** Tests pass. More importantly, manual inspection shows that you did a tremendous job here. I grew up in the modem-driven pre-Internet 80's Bulletin Board System (BBS) era where l33t ANSII art was an everyday necessity. Seeing ANSII codes embedded in the @beartype codebase sparks so much childlike joy for me. In other words, we are immediately merging this.

## That Said...

I'd like to do a bit of follow-up work on this. So, I guess `beartype 0.11.0` isn't quite as mission-complete as I thought it was. :face_exhaling: 

Specifically, **@beartype should only conditionally emit ANSII colour codes if `stdout` is attached to a non-monochrome interactive terminal that supports ANSII colour codes.** You are now thinking: "That's impossible to test for, bro." Hah! I knew you were thinking that. Thankfully, testing for that is not only possible but already solved. In order:

1. We at least need to test for whether `stdout` is attached to an interactive terminal. Thankfully, Python has a trivial one-liner idiom for that, too:

```python
from sys import stdout

def is_stdout_terminal() -> bool:
    '''
    ``True`` only if standard output is currently attached
    to an interactive terminal.

    See also this StackOverflow thread:
        https://stackoverflow.com/questions/3818511/how-to-tell-if-python-script-is-being-run-in-a-terminal-or-via-gui
    '''

    # One-liners for great justice.
    #
    # Note that input streams are *NOT* guaranteed to define
    # the isatty() method. Defensively test for the existence of
    # that method before deferring to that method.
    return hasattr(stdout, 'isatty') and stdout.isatty()
```

2. We then need to test for whether that terminal supports ANSII colour codes. Django, which I trust to get this right, supports this detection via [a pretty hardcore `django.core.management.color.supports_color()` tester function](https://github.com/django/django/blob/fdf0f625216cc5a70d28a3ac9a41f41935f1827c/django/core/management/color.py#L21). If you deeply examine that function, though, you'll find that function trivially reduces to the above `is_stdin_terminal()` function on POSIX-compliant platforms. So, everybody except Windows. Detecting ANSII colour support under Windows is *totally* insane, however. Let's not do that. Instead, let's simply unconditionally disable ANSII colour support under Windows. In short, this is the final detection algorithm I'm thinking of:

```python
from sys import platform, stdout

def is_stdout_terminal_colourized() -> bool:
    '''
    ``True`` only if standard output is attached to an
    interactive terminal that supports ANSII colour codes.
    '''

    # *BOOM.*
    return is_stdout_terminal() and platform != 'win32'


def is_stdout_terminal() -> bool:
    '''
    ``True`` only if standard output is currently attached
    to an interactive terminal.

    See also this StackOverflow thread:
        https://stackoverflow.com/questions/3818511/how-to-tell-if-python-script-is-being-run-in-a-terminal-or-via-gui
    '''

    # One-liners for great justice.
    #
    # Note that input streams are *NOT* guaranteed to define
    # the isatty() method. Defensively test for the existence of
    # that method before deferring to that method.
    return hasattr(stdout, 'isatty') and stdout.isatty()
```

Works for me. :relieved: 

If you find a free moment this weekend, would you mind submitting a follow-up PR that conditionally embeds ANSII colour codes in @beartype violations *only* if `is_stdout_terminal_colourized()` is `True`? If not, I'll happily make that happen myself this weekend. But... you know. You're clearly amazing and very much out of my league. Thought you might like to tackle this one, if you'd like.

## In Conclusion

**You're stunning!** Let's merge this immediately before I start playing video games compulsively.

leycec:
**Urgh!** [Tests that I have just enabled are sadly failing](https://github.com/beartype/beartype/runs/8151125960?check_suite_focus=true). Why? Because our test suite performs regex-based searches on @beartype violation strings to ensure that we're printing what we think we're printing. Of course, our regexes know nothing about ANSII colour codes. Mayhem ensues.

It looks like we'll also need to patch up our `beartype_test/a00_unit/a90_decor/a60_main/test_codemain.py` submodule to strip all ANSII colour codes from @beartype violations *before* performing regex-based searches on those violations. Although I'd be happy to handle that, you might prefer to have a go at that first?

Either way, let's hold off on merging this for the moment. I know, I know. It is sad. :sob: 

This is still ultra-cool, though. This is *almost* working perfectly. Let's see if we can gently push this across the finish line in time for `beartype 0.11.0`.

justinchuby:
Thanks for the detailed suggestions! I am going to update the PR hopefully today or tomorrow. 

You can help me with your beartype creator knowledge:

1. I am hoping to color the correctly matched type green (*List* index 0 ...). 
  Do you think that's a good idea? If so, how can I do that?
2. I wasn't able to find the code that outputs the first line `Not <class ...> or ...` (in the issue) to color the types as truth. Could you point me to where I should edit?
3. What do you think of the color scheme. Is there anything I should change?
4. Both the Windows terminal and VS Code support ANSI colors now. That said I am happy to exclude windows for now and see if there are better solutions.

justinchuby:
I updated the tests. Please feel very free to merge if you would like to explore some of the ideas above outside of this PR!

justinchuby:
> ** You are now thinking: "That's impossible to test for, bro." Hah! I knew you were thinking that.

You read my mind _and_ provided a solution. I can only assume you know magic!

leycec:
**Yar!** Tests be passing, which is wondrous. All hail @justinchuby! Eternal glory to our Foxy Overlord.

That said, [@langfield had the right of it with respect to backticks:](https://github.com/beartype/beartype/issues/161#issuecomment-1234708124) they genuinely do present an existential threat, albeit not quite for the reasons @langfield suggested. Yes, yes – \`\`\`-fenced code blocks in GitHub-flavoured MarkDown (GFMD) exist. But that's not the *real* problem. The *real* problem that could conceivably result in the worst-case horror show of a Big Tech Giant (think Microsoft here) submitting a  [Common Vulnerabilities and Exposures (CVE)](https://cve.mitre.org) against @beartype is that...

drumroll please

...the backtick is an operator in most POSIX-compatible shells (e.g., `bash`, `zsh`) referred to as the **command substitution operator.** So, what happens when somebody malicious gets somebody unintelligent to copy-and-paste a @beartype violation message like the following into their shell prompt?

```bash
$ @beartyped __main__.quote_wiggum() return `'remove_filesystem'` violates type hint `None`, as string `'remove_filesystem'` not instance of <class "builtins.NoneType">.
Removing filesystem...    # <--- wait, what? no, no! <ctrl-c>, <ctrl-c>!!!!!!!
Removed! Have a great day.   # <--- @beartype, i now hate you.
```

And... now @beartype is in shell injection attack territory. The problem isn't necessarily that we're outputting the backtick character – although that's always concerning. The problem is that we're outputting the backtick character **with arbitrary user-defined strings.** Let's not output the backtick character with arbitrary user-defined strings, people.

leycec:
I was *super*-hoping to merge and push this out with `beartype 0.11.0` tonight, but [backtick security concerns](https://github.com/beartype/beartype/pull/162#issuecomment-1236056912) kinda creep me out. Let's let this simmer a bit longer.

Thankfully, @justinchuby asks the essential questions above. Break time! *Break it down.*

> I am hoping to color the correctly matched type green (List index 0 ...). Do you think that's a good idea? If so, how can I do that?

**Sure!** Sounds great. Uhm... let's see here. *sifts through hundreds of thousands of lines of obfuscated code my younger self really thought was a good idea that now cause me to reconsider the wisdom of my younger self*

It looks like the exact substring in question here (i.e., `"List index 0"`) is emitted by the `beartype._util.text.utiltextlabel.label_obj_type()` utility function internally called by our ~~stupidly~~ cleverly deeply nested hidden `beartype.beartype._decor._error._pep._pep484585._errorsequence` submodule. Look. Don't ask. These things happen when you're hopped up on sleep deprivation, existential malaise, and a burning desire to bludgeon the nearest ergonomic keyboard with a filth-encrusted toothbrush.

> I wasn't able to find the code that outputs the first line `* Not <class ...> or ...` (in the issue) to color the types as truth. Could you point me to where I should edit?

:face_exhaling: 

I... uhm... err... yes. Of course. Of course, I know exactly where that happens. *furious `grep` action*

**Okay.** This is sorta super-magic right here. Basically, the ~~stupidly~~ cleverly deeply nested hidden `beartype.beartype._decor._error._pep._pep484._errorunion` submodule constructs a human-readable substring describing:

* The proper subset of the passed union type hint that superficially violates the current object being type-checked, where "superficially violates" means "has the wrong type."
* The proper subset of that hint that deeply violates the current object being type-checked, where "deeply violates" means "has the right type but contains one or more items that violate something."

If I'm reading you right, you'd like to theme the former. The "superficial violation" subset, right?

**Okay.** So, the `get_cause_or_none_union()` getter internally differentiates between these two types of violations by separating them into discrete local sets. The `hint_classes_unsatisfied` collects all union child type hints that are superficially violated, which is ultimately responsible for emitting the `* Not <class ...> or ...` substring that you're referencing here. I suspect that this `if` conditional in that getter is what you've been searching for with deep longing:

```python
def get_cause_or_none_union(sleuth: CauseSleuth) -> Optional[str]:
    ...

    # If this pith fails to shallowly satisfy one or more classes, concatenate
    # these failures onto a discrete bullet-prefixed line.
    if hint_classes_unsatisfied:
        # Human-readable comma-delimited disjunction of the names of these
        # classes (e.g., "bool, float, int, or str").
        cause_types_unsatisfied = join_delimited_disjunction_types(
            hint_classes_unsatisfied)

        # Prepend this cause as a discrete bullet-prefixed line.
        #
        # Note that this cause is intentionally prependend rather than appended
        # to this list. Since this cause applies *ONLY* to the shallow type of
        # the current pith rather than any items contained in this pith,
        # listing this shallow cause *BEFORE* other deeper causes typically
        # applying to items contained in this pith produces substantially more
        # human-readable exception messages: e.g.,
        #     # This reads well.
        #     @beartyped pep_hinted() parameter pep_hinted_param=(1,) violates
        #     PEP type hint typing.Union[int, typing.Sequence[str]], as (1,):
        #     * Not int.
        #     * Tuple item 0 value "1" not str.
        #
        #     # This does not.
        #     @beartyped pep_hinted() parameter pep_hinted_param=(1,) violates
        #     PEP type hint typing.Union[int, typing.Sequence[str]], as (1,):
        #     * Tuple item 0 value "1" not str.
        #     * Not int.
        #
        # Note that prepending to lists is an O(n) operation, but that this
        # cost is negligible in this case both due to the negligible number of
        # child hints of the average "typing.Union" in general *AND* due to the
        # fact that this function is only called when a catastrophic type-check
        # failure has already occurred.
        causes_union.insert(0, f'not {cause_types_unsatisfied}')
```

Basically, `beartype._util.text.utiltextjoin.join_delimited_disjunction_types()` is what I'm saying. Gah! Why is nothing easy when @beartype is involved? And the answer is... because my younger self enjoyed an excess of child-like enthusiasm coupled with a deficit of adult-like wisdom.

> What do you think of the color scheme. Is there anything I should change?

**Love it!** @langfield, what say ye, eldritch wise sage of the old ANSII ways?

> Both the Windows terminal and VS Code support ANSI colors now. That said I am happy to exclude windows for now and see if there are better solutions.

You're... almost certainly right. Instinctively, I don't trust the pre-2020 Microsoft to get POSIX-compliance right. But Microsoft is a newly revived beast that grows smarter with each passing day. Perhaps they've finally gotten ANSII colours right?

I choose to believe the words that you're saying, @justinchuby. I have only one question: "Are there any popular Windows terminals that @beartype might be run in that still do *not* support ANSII?" I'm not fussed about decrepit backward-compatible Windows terminals that nobody cares about anymore. I'm mostly just concerned about dark corner cases that will cause @beartype users to erupt with wrathful anger and insensitive justice-seeking on our issue tracker. Because those people scare me.

justinchuby:
I didn't know about command substitution. Thanks, a lesson learned!

How do we feel about just coloring; quotation marks `'`; `-> b'some value' <-`; `👉 emoji👈`?

justinchuby:
image

justinchuby:
> Are there any popular Windows terminals that https://github.com/beartype might be run in that still do not support ANSII?

That's a great question. From windows 11 Windows Terminal has become the default. But I think pre windows 10 systems are still running cmd in the old Windows Console

- [x] VS Code
- [x] [Windows Terminal](https://github.com/microsoft/terminal)
- [?] Visual studio (Need to check)
- [x] JetBrains IDEs
- [ ] 👎 Windows Console

leycec:
**Phenomenal achievement.** This is the ultimate hotness. I'm incredibly impressed by the prowess you demonstrated here. You've done a tremendous act of charity and kindness for bear bros the world over. I wish you the utmost success with your own cutting-edge research on the periphery of ML. You're stunning.

We're merging this beautification post-haste. A bold new colourful future beckons.

> I changed colors to distinguish user inputs and the types of those inputs.

:open_mouth: 

> ...quotation marks 'b'some value'' «b'some value'» 「b'some value'」; -> b'some value' <-; emoji 👉 b'some value' 👈?

Superb question! Unicode code points are, of course, aesthetically pleasing and hot. I myself rock ultra-customized `zsh` prompts and Vim configurations plastered with excessive Unicode like a 70's disco inferno.

I'm unclear whether Python can portably detect support for Unicode code points in the current TTY font(s), though. The portability concern is this: if the font(s) configured for the current TTY do *not* support the Unicode code points we are trying to print, then the TTY will uselessly print □ (i.e., the placeholder Unicode code point) instead. I should probably do a bit of research on this. This is 2022. Right? There's *got* to be a way to portably detect Unicode TTY support in 2022. *...doesn't there?* :face_exhaling: 

> From windows 11 Windows Terminal has become the default.

I'm convinced! Since pre-Windows 10 systems are a compelling security risk, we can safely ignore them. Anyone on pre-Windows 10 has more problems than improperly formatted ANSII colours.

I'll go ahead and remove all `sys.platform() != 'win32'` checks after merging this tonight. Thanks again for such a tremendous outpouring of open-source enthusiasm and intellect. You rock everything. :guitar: :notes:

beartype_test/a00_unit/a90_decor/a60_main/test_codemain.py
```diff
@@ -12,13 +12,22 @@
...
 #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 from beartype.roar import BeartypeDecorHintPep585DeprecationWarning
 from beartype_test._util.mark.pytmark import ignore_warnings
 
+
+def _strip_ansi_escape_sequences(text: str) -> str:
+    '''
+    Strip all ANSI escape sequences from the passed string.
+    '''
+    return re.sub(r'\\033[^m]*m', '', text)
```
justinchuby, line 28: @leycec i realized I may have made a mistake here. Should it be `\033` without the extra `\`? Or does it need to be in hex format? When I did this I thought I needed to match the literal `\033`. Sorry this really is my first time working with escape sequences. I can run a test today. 

The tests passed because the colors were not added in the environment so there's nothing to match for.

justinchuby: Looks like either way works 🤔

leycec: **Fascinating.** You're... right. I'm not quite sure, honestly. The intersection of ANSII escape sequences with raw regex syntax is surprisingly non-trivial. Thanks a heap for investigating that!

Luckily for us, it looks like you already nailed it on the first try. Just in case something actually *is* awry here, I've added an internal `FIXME:` comment to that line suggesting further research.

As always, you amaze. :bow:

leycec:
Thanks to your gruelling market research, commit 5d6fca01972fe extends colourization support to Windows! What @leycec promised, @leycec actually delivered. :smile: 

Let's think about how to detect Unicode terminal support at some point, too. That's hot stuff. Until then, we feel profound gratitude for everything you've done here. May you win all the grant funding. :money_mouth_face:

justinchuby:
Thanks @leycec for your very kind words. It makes me feel that I have done much more than I did. Thank YOU for creating the fantastic @beartype!

PR #177, opened 2022-10-06
Expose beartype's `__all__`

pinkwah:
Mypy is unhappy because beartype isn't explicitly specifying its exported attributes. I assume this is a mistake since this code is not deleted but commented out.

```
error: Module "beartype" does not explicitly export attribute "beartype"; implicit reexport disabled
```

### DIFF
--- a/beartype/__init__.py
+++ b/beartype/__init__.py
@@ -91,13 +91,13 @@ For PEP 8 compliance, this specifier has the canonical name
 '''
 
 
-# __all__ = [
-#     'BeartypeConf',
-#     'BeartypeStrategy',
-#     'beartype',
-#     '__version__',
-#     '__version_info__',
-# ]
+__all__ = [
+    'BeartypeConf',
+    'BeartypeStrategy',
+    'beartype',
+    '__version__',
+    '__version_info__',
+]
 '''
 Special list global of the unqualified names of all public package attributes
 explicitly exported by and thus safely importable from this package.


leycec:
*...omg.* Thanks so much for your fast-action fix, @pinkwah. As expected, you're totally right. This was a gross API oversight on our part. Let's merge this immediately for the betterment of all QA kind.

Also, may I say what a lovely apple-eating avatar you have there! My wife and I have a handful of dwarf zone-hardy apple trees dotted about our wilderness property. Because they are as lazy as I am, they have yet to actually produce any apples. I shake my fist every time I walk past one. :apple: :fist_raised: :apple:

pinkwah:
It happens!

And thanks! My aunt has some apple trees but they're also too lazy, even though they get apples. Turns out reindeer _love_ apples that have been on the ground fermenting. They woke up to three of them passed out drunk in their garden.

leycec:
Hah, hah... *hah.* Oh, @pinkwah. That's a timeless Norwegian moment for the ages. Piss-drunk reindeer hopped up on rotting cider apples is the memory I will cherish all week and repeat to my wife until she finally orders me to stop in a fit of exasperation.

We don't have reindeer stories in Ontario. I wish we did. I would spam GitHub issues and PRs like this non-stop with piss-drunk reindeer pics if we did, because I would personally feed them rotting cider apples.

Instead, we have *raccoon* stories. Raccoons are like reindeer – only smaller, deadlier, and more inclined to eat everything not already covered in cat fur. This is our story.

One morning, my wife and our Bengal cat were drinking coffee together on our lakeside dock – just like they always used to. The summer Sun beat warmly on their backs. The lake was quiet with the susurration of crickets. Suddenly, the Bengal growls like Hell itself is about to be unleashed upon the unrighteous. My wife turns.

A raccoon juiced up on "edible forest mushrooms" is running full-tilt down the hill from our house to the dock. My wife screams and jumps five feet. The Bengal howls and dashes for cover under a canoe. The raccoon blindly barrels past both, leaps with a furious squeal at the edge of the dock, and cannonballs directly into the lake. Water erupts everywhere. My wife is still screaming. The Bengal is still howling.

The Bengal never sat with my wife on the dock again.

PR #179, opened 2022-10-12
re-add a mypy github action step, using the local python version and env

jpetrucciani:
This PR should allow this action to use mypy to validate PEP 561, without using any external actions that may not support the OSes in our matrix!

related:
https://github.com/jpetrucciani/mypy-check/issues/30
https://github.com/jpetrucciani/mypy-check/issues/31

### DIFF
--- a/.github/workflows/python_test.yml
+++ b/.github/workflows/python_test.yml
@@ -234,36 +234,15 @@ jobs:
         with:
           python-version: "${{ matrix.python-version }}"
 
-      #FIXME: Consider reenabling this action at some later date – or perhaps
-      #not. I threw more than a few scarce volunteer hours at this action with
-      #little to no success but great agitation and increasingly wrathful
-      #feelings. The final straw was a seemingly non-human-readable
-      #Windows-specific error at action setup resembling:
-      #    unable to prepare context: path "D:a_actionsjpetruccianimypy-checkmaster" not found
+      # Type-check this package with "mypy".
       #
-      #In short, we're done here for now. Rather than depend on an external
-      #action for this, it would probably be prudent to simply:
-      #* Install "mypy" as a standard Python package.
-      #* Run "mypy" as a standard CLI command.
-      #
-      #In short, that's what we *SHOULD* have done initially. More fool us.
-
-      # Type-check this package with "mypy". See also:
-      #     https://github.com/jpetrucciani/mypy-check
-      #
-      # Note that:
-      # * This otherwise excellent action currently requires Docker and thus
-      #   fails to support macOS runners, due to a petty licensing dispute on
-      #   Docker's part. Upstream wisely advises skipping this action under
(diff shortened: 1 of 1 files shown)


leycec:
**Brilliance personified.** I now feel super-bad. You've gone above and beyond the call of open-source duty to repair our  badly done CI workflow to support `mypy` *without* even leveraging the [GitHub Action you invested so much volunteer time and energy in](https://github.com/jpetrucciani/mypy-check).

I'm extremely grateful. I'm also extremely apologetic for unnecessarily acting out on your issue tracker. It's been a rough real-world week for the @leycec family. That does *not* excuse my rudeness, of course; it just contextualizes it. :sweat: 

You've been so awesome here and elsewhere! I'll happily revisit using `jpetrucciani/mypy-check` once the dust settles. Until then, thanks so much for this extraordinarily generous contribution to the @beartype clan. So much :heart_eyes_cat:.

jpetrucciani:
No need to feel bad! 😄 I believe that helping out where I am able is one of the coolest and most fulfilling parts of open source 🚀 

Completely understand (and it is no big deal!) - I hope that the real-world takes it a bit easier on you and your family soon!

as for `jpetrucciani/mypy-check` - don't worry about revisiting it in the future - this step should handle everything you need, and will be easy to update and tweak as you need! my action was much more useful in the earlier days of GitHub actions where environments were a bit more difficult to get into place (so it was a short-hand, sandboxed way of running mypy without any additional setup). It still works pretty well (albeit not the new optimal way of doing it!), but as you've discovered, that only seems to be the case for Linux based systems 😞

leycec:
**Oh!** That makes *perfect* sense.

Because I am ancient and crusty, I stumbled years later into this bustling GitHub Actions scene. I didn't even know that there had once been a Bad Ol' Days™ in which Python environments were non-trivial to setup and maintain. Pinch me, I must be dreaming. We're now living in the Good New Days™, everyone!

But there's *absolutely* still a place for thoughtful actions like yours. Indeed, I'd love to see a sort of Mega Static Type-checking Action that transparently supports *all* existing well-maintained static type-checkers for Python – including:

* **Mypy,** naturally.
* ~~FaceBook's~~ Meta's [Pyre](https://pyre-check.org).
* Google's [`pytype`](https://github.com/google/pytype).
* Microsoft's [`pyright`](https://github.com/Microsoft/pyright).

Since mypy is Python's official static type-checker, that action would type-check via mypy by default. For usability, that action would also transparently support all of Python's *other* static type-checkers with a `type-checkers` list setting resembling:

```yaml
      - name: 'Type-checking package with *LITERALLY EVERYTHING*...'
        uses: 'jpetrucciani/python-checkup@v1'
        with:
          python-version: "${{ matrix.python-version }}"
          type-checkers: ['mypy', 'pyright']
```

And... I checked. `python-checkup` is free. :wink: :stuck_out_tongue_winking_eye:

PR #214, opened 2023-03-09
[Documentation] Clarified comments on linecache module

faangbait:
Existing comment made this sound like a hack. This clarifies that it's a supported value intended to no-op.


leycec:
**Superb!** Thanks so much for the solid clarification, @faangbait. You're absolutely right, of course; that's non-trivial code I barely understand myself. In fact, I'm pretty sure this is a relic of @posita's mad genius. <sup>*You hear that, @posita? You're a mad genius now.*</sup>

I honestly didn't even know what ``None`` meant in this context. Tests passed, time and baldness proceeded, and we just pretended we knew what was happening. When I can't make it, I fake it. :wink:

leycec:
**It's in.** Also, I love the "Valentine's Day Edition" Daft Punk avatar! I would totally trust that robot when it comes for me. It only wants to help, after all. :robot:

PR #266, opened 2023-07-24
Update FAQ to recommend jaxtyping even for PyTorch

patrick-kidger:
I'd actually really recommend jaxtyping for NumPy annotations as well, but that's perhaps mildly controversial, given the existence of `nptyping`, and the official NumPy typing hints. I can do a separate PR for that if desired.

### DIFF
--- a/doc/src/_links.rst
+++ b/doc/src/_links.rst
@@ -843,8 +843,6 @@
    https://github.com/google/jaxtyping
 .. _nptyping:
    https://github.com/ramonhagenaars/nptyping
-.. _TorchTyping:
-   https://github.com/patrick-kidger/torchtyping
 
 .. # ------------------( LINKS ~ soft : ide                  )------------------
 .. _PyCharm:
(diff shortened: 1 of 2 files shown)


leycec:
**Wunderbar!** Thanks so much for this superb PR, which resolves long-standing issue #194 and relieves me of a considerable burden that I was not likely to resolve myself. Let's merge this immediately for the good of QA- and AI-kind.

Relatedly...

> I'd actually really recommend `jaxtyping` for NumPy annotations as well, but that's perhaps mildly controversial, given the existence of `nptyping`, and the official NumPy typing hints. I can do a separate PR for that if desired.

*Yes! So much yes!* When you find a spare volunteer second, a second PR briefly highlighting the tradeoffs between these three related APIs would be just supernal. I myself am unclear about those tradeoffs – but suspect that `jaxtyping` easily edges out the other two, because of course it does. Also, I am lazy.

This is why I leave everything in your capable hands, Dr. Kidger. :wink:

Issue #302, opened 2023-11-07
[Feature Request] New `BeartypeConf(hint_mappings)` option. @felixchenier deserves it!


felixchenier:
Hello hello

Here is me, coming with bad numpy stuff.

This code creates a numpy array of integers, then indexes one of these integers to be printed by a function that accepts integers as an input. End of the story.

```python
import numpy as np
from beartype import beartype, BeartypeConf

typecheck = beartype(conf=BeartypeConf(is_pep484_tower=True))


@typecheck
# Functions
def print_integer(var: int):
    print(var)


# Stuff
if __name__ == "__main__":
    table = np.array([1, 1])
    i = table[0]
    print(type(i))
    
    print_integer(i)
```

But to my grandiose sadness, numpy defines its own integer class and everything falls apart.

```python
<class 'numpy.int64'>
Traceback (most recent call last):

  File ~/opt/miniconda3/envs/mosa/lib/python3.10/site-packages/spyder_kernels/py3compat.py:356 in compat_exec
    exec(code, globals, locals)

  File ~/Documents/git/kineticstoolkit/kineticstoolkit/untitled0.py:19
    print_integer(i)

  File <@beartype(__main__.print_integer) at 0x15e519990>:21 in print_integer

BeartypeCallHintParamViolation: Function __main__.print_integer() parameter var=1 violates type hint <class 'int'> under non-default configuration BeartypeConf(claw_is_pep526=True, is_color=None, is_debug=False, is_pep484_tower=True, strategy=<BeartypeStrategy.O1: 2>, warning_cls_on_decorator_exception=None), as <protocol "numpy.int64"> 1 not instance of int.
```

So beartype does the correct thing, but the correct thing is not correct. To whom should I address my prayers?

leycec:
@posita: I politely summon thee to solve this desperate plea from a very nice man who has also submitted a [very nice PR](https://github.com/beartype/beartype/pull/301). In the hopes that this man will continue to submit very nice PRs, I wish to appease this man. I probably shouldn't have written exactly what I was thinking. But that's what happened anyway.

Thankfully, I *think* I already know what *probably* needs to be done here. Here's my uneducated **tl;dr**. To transparently support NumPy scalars under runtime type-checkers like @beartype, please do:

[...]

PR #309, opened 2023-11-24
Feature 302 done, but not.

felixchenier:
Hi @leycec

Feature #302 is done. But not.

Having a dict parameter seems to be problematic because the current code tries to build a hashable tuple of parameters (confcls.py, line 498 of this commit) but a dict is not hashable. Therefore it strongly complains.

Unfortunately I could not do much more work on it, because the whole configuration caching procedure is way out of my understanding. At least I added "stuff" so that the new config parameter now exists and it "should work". 😕

Sorry to throw this back to you!

### DIFF
--- a/beartype/_check/convert/convreduce.py
+++ b/beartype/_check/convert/convreduce.py
@@ -267,6 +267,10 @@ def _reduce_hint_cached(
         * If the passed hint is reducible, another hint reduced from this hint.
         * Else, this hint as is unmodified.
     '''
+    # If this configuration maps this hint to another hint, do so now.
+    # This one-liner looks ridiculous, but actually works. More
+    # importantly, this is the fastest way to accomplish this. Flex!
+    hint = conf.hint_overrides.get(hint, hint)    
 
     # Sign uniquely identifying this hint if this hint is identifiable *OR*
     # "None" otherwise.
--- a/beartype/_conf/_conffrozendict.py
+++ b/beartype/_conf/_conffrozendict.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python3
+# --------------------( LICENSE                            )--------------------
+# Copyright (c) 2014-2023 Beartype authors.
+# See "LICENSE" for further details.
+
+'''
+Beartype frozen dict, used to memoize configuration options such as
+hint_overrides.
+'''
+
+class _BeartypeFrozenDict(dict):
+    '''
+    **Frozen dictionary** (i.e., immutable mapping preserving :math:`O(1)`
+    complexity while prohibiting modification).
+
+    Instances of this dictionary are safely hashable and thus suitable for
+    passing as parameters to memoized callables and classes (e.g., our core
+    :class:`beartype.BeartypeConf` class).
+    '''
+
+    __slots__ = ('_hash',)  # <-- preserve space efficiency. curse you, Python!
+
+    def __init__(self, *args, **kwargs) -> None:
+
+        # Instantiate this immutable dictionary with all passed parameters.
+        super().__init__(*args, **kwargs)
+
+        # Precompute the hash for this immutable dictionary at instantiation
+        # time for both efficiency and safety.
+        frozen_items = frozenset(self.items())  # <-- clever stuff
+        self._hash = hash(frozen_items)  # <---- more clever stuff
+
+
+    def __hash__(self) -> int:
+        return self._hash
+
+
+    def __setitem__(self, key, value) -> None:
+        raise NotImplementedError(
+            f'Immutable dictionary {repr(self)}" '
+            f'key {repr(key)} not settable to {repr(value)}.'
+        )
--- a/beartype/_conf/confcls.py
+++ b/beartype/_conf/confcls.py
@@ -50,6 +50,7 @@ from beartype._data.hint.datahinttyping import (
     TypeException,
     TypeWarning,
 )
+from beartype._conf._conffrozendict import _BeartypeFrozenDict as BeartypeHintOverrides
 from beartype._data.func.datafuncarg import ARG_VALUE_UNPASSED
 from beartype._util.cls.utilclstest import is_type_subclass
 from threading import Lock
@@ -78,7 +79,7 @@ class BeartypeConf(object):
         from the names of all keyword parameters accepted by the :meth:`__new__`
         method to the corresponding values of those parameters in this
         configuration).
-    _hint_overrides : Dict[type, type]
+    _hint_overrides : BeartypeHintOverrides
         Instructs beartype to evaluate given types as other types. This
         is useful for defining a strict API but accepting a looser runtime
         type evaluation. For instance, calling `def function(param : int)`
@@ -177,7 +178,7 @@ class BeartypeConf(object):
         _claw_is_pep526: bool
         _conf_args: tuple
         _conf_kwargs: Dict[str, object]
-        _hint_overrides: Dict[type, type]
+        _hint_overrides: BeartypeHintOverrides
         _is_color: Optional[bool]
         _is_debug: bool
         _is_pep484_tower: bool
@@ -204,11 +205,11 @@ class BeartypeConf(object):
         # Optional keyword-only parameters.
         *,
 
-        #FIXME: Uncomment us when implementing O(n) type-checking, please.
+        # Uncomment us when implementing O(n) type-checking, please.
         # check_time_max_multiplier: Union[int, None] = 1000,
 
         claw_is_pep526: bool = True,
-        hint_overrides: Dict[type, type] = dict(),
+        hint_overrides: BeartypeHintOverrides = BeartypeHintOverrides(),
         is_color: BoolTristateUnpassable = ARG_VALUE_UNPASSED,
         is_debug: bool = False,
         is_pep484_tower: bool = False,
@@ -310,14 +311,14 @@ class BeartypeConf(object):
             performance-sensitive modules *after* profiling those modules to
             suffer performance regressions under import hooks published by the
             :mod:`beartype.claw` subpackage. Defaults to :data:`True`.
-        hint_overrides : Dict[type, type]
-            Instructs beartype to evaluate given types as other types. This
-            is useful for defining a strict API but accepting a looser runtime
-            type evaluation. For instance, calling `def function(param : int)`
-            would fail by default if param is a `numpy.int64`, which does not
-            inherit from `int`. By optionally setting
-            `hint_overrides={int: numbers.Integral}`,
-            the same call would not fail because beartype now considers any
+        hint_overrides : BeartypeHintOverrides
+            A dict that instructs beartype to evaluate given types as other
+            types. This is useful for defining a strict API but accepting a
+            looser runtime type evaluation. For instance, calling
+            `def function(param : int)` would fail by default if param is a
+            `numpy.int64`, which does not inherit from `int`. By optionally
+            setting `hint_overrides={int: numbers.Integral}`, the same call
+            would not fail because beartype now considers any
             `numbers.Integrals` as being an `int`.
         is_color : BoolTristateUnpassable
             Tri-state boolean governing how and whether beartype colours
@@ -495,10 +496,15 @@ class BeartypeConf(object):
             # value of the ${BEARTYPE_IS_COLOR} environment variable (if set).
             is_color = get_is_color(is_color)
 
+            # Convert hintoverrides from dict to BeartypeHintOverrides
+            #FIXME: Currently type checking is done after this conversion.
+            #Therefore this conversion may fail before being checked.
+            hint_overrides_as_BeartypeHintOverrides = BeartypeHintOverrides(hint_overrides)
+
             # Efficiently hashable tuple of these parameters in arbitrary order.
             conf_args = (
                 claw_is_pep526,
-                hint_overrides,
+                hint_overrides_as_BeartypeHintOverrides,
                 is_color,
                 is_debug,
                 is_pep484_tower,
@@ -529,7 +535,7 @@ class BeartypeConf(object):
             #
             # If "hint_overrides" is *NOT* dict of types:types, raise an exception.
             elif (
-                not isinstance(hint_overrides, Dict)
+                not isinstance(hint_overrides_as_BeartypeHintOverrides, BeartypeHintOverrides)
                 or not all(isinstance(_, type) for _ in hint_overrides.keys())
                 or not all(isinstance(_, type) for _ in hint_overrides.values())
             ):
--- a/beartype/_conf/confcls.py
+++ b/beartype/_conf/confcls.py
@@ -79,7 +79,7 @@ class BeartypeConf(object):
         from the names of all keyword parameters accepted by the :meth:`__new__`
         method to the corresponding values of those parameters in this
         configuration).
-    _hint_overrides : Dict[type, type]
+    _hint_overrides : Dict
         Instructs beartype to evaluate given types as other types. This
         is useful for defining a strict API but accepting a looser runtime
         type evaluation. For instance, calling `def function(param : int)`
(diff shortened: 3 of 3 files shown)


leycec:
> a dict is not hashable

**GAH.** Woopsie. I... sorta forgot all about that roadblock in the way of glorious progress. Caching is pretty integral to this whole `BeartypeConf` thing. This issue is entirely my fault, @felixchenier. I apologize profusely for neglecting this fundamental.

My apology is profuse! :sweat: 

## Kk. We Still Got This...

**Fear not.** I have a question for you, @felixchenier. The fate of our world may depend upon your answer:

> Have you ever wanted to implement a frozen dictionary data structure in pure-Python that preserves efficiency, immutability, maintainability, and readability? Moreover, have you ever wanted to do so in **just 12 lines of code ignoring docstrings, comments, and blank lines**? Well!? HAVE YOU!?!?

*...oh.* You never wanted to do something like that because you value your very soul, huh? That's fair enough. Thankfully, I sell my soul more cheaply than one of our cats for stinky salmon meat. So, I did this for everybody. It's best not to ask why people like us do these sorts of things at 1:23AM on a bitterly chill Friday night. It's gotta be for more than just the lulz! <sup>**Spoiler:** *it's just for the lulz.*</sup>

```python
# Please put this private data structure somewhere, Felix. May Guido save our souls.
class _BeartypeFrozenDict(dict):
    '''
    **Frozen dictionary** (i.e., immutable mapping preserving :math:`O(1)`
    complexity while prohibiting modification).

    Instances of this dictionary are safely hashable and thus suitable for
    passing as parameters to memoized callables and classes (e.g., our core
    :class:`beartype.BeartypeConf` class).
    '''

    def __init__(self, *args, **kwargs) -> None:

        # Instantiate this immutable dictionary with all passed parameters.
        super().__init__(*args, **kwargs)

        # Precompute the hash for this immutable dictionary at instantiation
        # time for both efficiency and safety.
        frozen_items = frozenset(self.items())  # <-- clever stuff
        self._hash = hash(frozen_items)  # <---- more clever stuff


    def __hash__(self) -> int:
        return self._hash


    def __setitem__(self, key, value) -> None:
        raise NotImplementedError(
            f'Immutable dictionary {repr(self)}" '
            f'key {repr(key)} not settable to {repr(value)}.'
        )
```

And for the *coup de grace*: <sup>See what I did there, bro? Pretty sure that was awesome French.</sup>

```python
# Please make this alias available for external public export by lovely users
# such as yourself as:
#     from beartype import BeartypeHintOverrides
BeartypeHintOverrides = _BeartypeFrozenDict
```

I even tested this monstrosity. It works! Oh, hallelujah! Our horrifying monster... *it moves.*

```python
# Test that dictionary lookup still works.
>>> heh = BeartypeHintOverrides({'lol': 0xFEEDFACE, 'wut': 0xCAFEBABE})
>>> print(heh['lol'])
4277009102

# Test that dictionary mutability now fails.
>>> heh['guhh'] = 'MUH MUTABILITY!!!!!!!!!!'
NotImplementedError: Immutable dictionary {'lol': 4277009102, 'wut': 3405691582}" key 'guhh' not settable to 'MUH MUTABILITY!!!!!!!!!!'.

# Test that this immutable dictionary can now be the key of a mutable dictionary.
>>> yum = {heh: 'we got this'}
>>> print(yum)
{{'lol': 4277009102, 'wut': 3405691582}: 'we got this'}  # <-- *OMG.* Madness erupts.
```

@beartype was gonna need something like this, sooner or later. I'm only sorry you had to be the one to see us like this. :sob:

felixchenier:
@leycec Can you hear the angels sign? Aaaaaaaaaaaaaaaaaaaa

This code works, and this rocks.

```
import beartype
import numbers
import numpy as np

typechecked = beartype.beartype(
    conf=beartype.BeartypeConf(
        hint_overrides={int: numbers.Integral, float: numbers.Real}
    )
)

@typechecked
def test(a: int, b: float):
    print(a, b)

test(1, 2.0)
# Passes. Obvious.

test(1, 2)
# Passes. Yeah!

test(1, np.array([1])[0])
# Passes. Yeah!

test(1.0, 2.0)
# Fails. Yeah!
```

OK now there are some things to say.

1. Mypy is not happy again, it does not seem to like your frozen dict. Maybe it likes it less frozen, a little bit warmer. Who are we to discuss its tastes.

2. I think I did something you won't like. Instead of requiring a BeartypeHintOverrides in config, I let it be a dict as it was supposed to be. It is converted to a frozen dict internally. In my view, the user is not supposed to be bothered by internal caching considerations. But it means that the saved configuration is not *exactly* the same as what we provided. To get around this, I also convert it back to a dict in the property's getter. So for the user, it's a dict. For beartype, it's a frozen dict. Are you okay with this ? If not, no problem, I will (reluctantly but happily - is it possible) change it to BeartypeHintOverrides everywhere.

3. Type checking the BeartypeConf arguments happens *after* creating the arguments tuple. Therefore, I needed to convert hint_overrides from a dict to a BeartypeHintOverrides *before* checking if it really was a dict. The common python user would say: "that's ok, it will yell if it doesn't work". On the other hand, the common @leycec would yell: "not ok, I say it doesn't work". Am I wrong?

4. I need to answer your question: 

> Have you ever wanted to implement a frozen dictionary data structure in pure-Python that preserves efficiency, immutability, maintainability, and readability? Moreover, have you ever wanted to do so in just 13 lines of code ignoring docstrings, comments, and blank lines? Well!? HAVE YOU!?!?

In reality, mmmmm, let me think... Ah, no. Sorry! 😉

Big thanks for letting me doing a very little nothing to make beartype cover 100% of my needs!

leycec:
**Fantasticness.** The fat naked cherub things did sing. What are those cherub things, anyway? *Sus af.*


<sup>do not trust this man</sup>

> Mypy is not happy again.

Mypy, you have failed us for the last time. Clearly, mypy is one pancake short of a full stack:

```python
beartype/_conf/_conffrozendict.py:34: error: Signature of "__hash__" incompatible with supertype "dict"  [override]
beartype/_conf/_conffrozendict.py:34: note:      Superclass:
beartype/_conf/_conffrozendict.py:34: note:          None
beartype/_conf/_conffrozendict.py:34: note:      Subclass:
beartype/_conf/_conffrozendict.py:34: note:          def __hash__(self) -> int
Found 1 error in 1 file (checked 275 source files)
```

This... *is nonsense.* This happens because the builtin `dict` type defines `__hash__` like so:

```python
class dict(object):
    ...

    __hash__ = None
```

Mypy should know better than to fall for that nonsense. Instead, mypy is dumb. <sup>_\*facepalm\*_</sup>

> I think I did something you won't like.

**Heresy!** Sadly, we no longer live in the Medieval Ages. Shrieking "Heresy!" and then ripping at my beard is apparently no longer socially acceptable. The more things improve, the more we miss the old ways.

> Instead of requiring a `BeartypeHintOverrides` in config, I let it be a `dict` as it was supposed to be.

*...heh.* Thought you might do something clever like that. Quebec's best and brightest shines bright tonight, indeed.

> In my view, the user is not supposed to be bothered by internal caching considerations.

**This is good.** @leycec approval intensifies.

> But it means that the saved configuration is not exactly the same as what we provided.

**This is possibly super-bad.** Or... possibly not. I shall ruminate on this deep matter of existentialist philosophy while stroking my beard and playing video games thoughtfully.

The core issue here, really, is *speed.* That's why we all @beartype, right? Well, that and our dashing good looks! Gotta look suave while we QA-stomp those chittering bugs. :hiking_boot: :bug: 

The current `BeartypeConf.__new__()` implementation is *fast,* because it caches the passed arguments as is. The new `BeartypeConf.__new__()` implementation is a bit less fast, because it now has to modify those arguments before caching those arguments. We're not supposed to care about efficiency in Python. But I care! I'm obsessed. I probably shouldn't be. It's best just to accept that Python is slow so that we can all move on with our lives and get something done. Sadly, like mypy, *I too am dumb.*

This also introduces subtle complications elsewhere – like `__repr__()` no longer reflecting the actual contents of the dictionary. "Meh. Meh," I say.

## You Have Failed to Convince Me and You Should Feel Bad.

I know! I know, right? You still don't want to do this. Don't worry. I wasn't gonna make you do this. I just wanted you to read my giant monologue as punishment. <sup>Kidding! Only kidding.</sup>

This is phenomenal as is. I *might* refactor this later. I might not. The will of @leycec is capricious and unpredictable, much like the polar vortex now pelting poor Ontario with a vicious slurry of freezing rain, mud snow, and icy slush. Hear my prayer, Santa: "All I want for Christmas, New Years, and every day for the next six months is a thick blanket of real snow so that my wife and I can ski like proper Canadians. Thank you. Please send our regards to Mrs. Claws."

## There Is a Final Issue, But I Gots This One for Ya.

You weren't wrong you suggested that recursion is an issue. Sure, we *have* already handled recursion in unions (e.g., `{float: int | float}`). But we haven't actually handled recursion anywhere else! Like, the following is guaranteed to raise a `RecursionError` from @beartype:

```python
@beartype(conf=BeartypeConf({str: list[str]})
```

You are now furiously cogitating: "Merde. Merde. What have we done? Oh, mon dieu. Mais qui sait."

But fear not, my swarthy friend. With the disgusting power of regular expressions, we can explicitly detect and prohibit recursive hint overrides like so:

```python
from re import (
    escape as re_escape,
    search as re_search,
)

class BeartypeConf(object):
    def __init__(...) -> None:
        # For each source and target hint override in the passed dictionary...
        for hint_override_src, hint_override_trg in hint_overrides.items():
            # Regular expression matching subscription-style recursion in this
            # hint override (e.g., 'str: list[str]').
            #
            # Note that:
            # * Beartype currently only supports union-style recursion (e.g.,
            #   "float: int | float").
            # * Regular expressions inevitably fail in edge cases (e.g.,
            #   "str: Literal['str']"). Thankfully, hint overrides *AND* these
            #   edge cases are sufficiently rare that we can conveniently ignore
            #   them until some GitHub pyromaniac inevitably complains and
            #   starts lighting dumpster fires on our issue tracker.
            hint_override_recursion_regex = (
                # An opening "[" delimeter followed by zero or more characters
                # that are *NOT* the corresponding closing "]" delimiter.
                r'\[[^]]*'
                # The machine-readable representation of this source override,
                # escaped to protect all regex-specific syntax in this
                # representation from being erroneously parsed as that syntax. 
                fr'{re_escape(repr(hint_override_src))}'
                # Zero or more characters that are *NOT* the corresponding
                # closing "]" delimiter followed by that delimiter.
                r'[^]]*\]'
            )

            # Match object if this hint override contains one or more instances
            # of subscription-style recursion *OR* "None" otherwise.
            hint_override_recursion = re_search(
                hint_override_recursion_regex, repr(hint_override_trg)

            # If this hint override contains one or more instances of
            # subscription-style recursion, raise an exception.
            if hint_override_recursion is not None:
                raise BeartypeConfException(
                    f'Recursive type hint override '
                    f'{repr(hint_override_src)}: {repr(hint_override_trg)} '
                    f'currently unsupported. Please complain loudly on the '
                    f'@beartype issue tracker if you feel that this is dumb.'
                )
```

There... what could be simpler? :vomiting_face: 

**Let's merge this quick before I break something.** @felixchenier, your will be done!

felixchenier:
@leycec Now the PR is closed, but I shall still answer some of your comments! Only for discussing.

> The core issue here, really, is speed. That's why we all beartype, right?

Mmmmmn, NO! :-) I want to beartype only to check types at runtime for providing useful error messages to my package's users. The only other good option is typeguard.

I chose beartype over typeguard for the worst possible reason: I looked at issue trackers and docs from both packages, and saw that typeguard uses a straightforward, well-understable approach that mainly using isinstance to do its magic, and is developed by a sane person who seems to value restricting the broadness of the package, so that it stays maintainable. On the other hand, I saw that beartype uses a fucked top approach exploded in a thousand directions to cover any possible or impossible case we may or may never encounter, supporting every single half-baked type hint syntax that's been supported or not in the entire Python history, adding features just for fun, while being maintained by some crazy bear-loving weirdo who's bedtime reading is a Github bug tracker. Of course I chose beartype!

> You Have Failed to Convince Me and You Should Feel Bad.

If only you knew how I don't! Go with what you want, really! It's none of my business. As long as it's documented and working, I'm absolutely fine with it. But just my last half-cent: do we really save time by constructing a BeartypeHintOverrides dictionary before calling the BeartypeConf constructor, rather than inside the BeartypeConf constructor? I'm absolutely no wizard as I don't know that intricate stuff, but to me if there's a difference, it's more negligible than I could ever think of caring. In any case, importing and using a BeartypeHintOverrides is more coherent with your other choices (e.g., using an enumeration instead of an integer), and only for it, it would make sense to also use that BeartypeHintOverrides class.

> This wouldn't work: @beartype(conf=BeartypeConf({str: list[str]})

But why should it? Isn't it a bit insane as it defeats the purpose of type hints? I agree that attempting this should cause an error. And congratulations for keeping you stomach contents inside while you were coding regular expressions. I never attempted to even approach that stuff, it seems so venomous.

Thanks for integrating this PR!

leycec:
> On the other hand, I saw that beartype uses a fucked top approach exploded in a thousand directions to cover any possible or impossible case we may or may never encounter,



> supporting every single half-baked type hint syntax that's been supported or not in the entire Python history, adding features just for fun, while being maintained by some crazy bear-loving weirdo who's bedtime reading is a Github bug tracker.



> Of course I chose beartype!



> But just my last half-cent: do we really save time...

**You have convinced me.** That's basically impossible... but you did it. Even though I feel old, am bald, and prefer pretending that other people aren't speaking when they are, you changed my mind.

**...is what I *would* be saying if literally everything wasn't broken!** But it was. Seriously, bro. Our test suite catastrophically broke after I optimistically merged this PR with my eyes wide shut. I :heart_eyes_cat: you so much – but I can't help but question whether you locally ran unit tests with `pytest` on your end? Cause... *it kinda seems like you didn't.*

On the bright side, I've now repaired everything that broke. On the dark side, doing so highlighted the fragility of trying to internally coerce the passed `hint_overrides` parameter from `dict` to `_BeartypeFrozenDict` in `BeartypeConf.__new__()`. Technically, doing so *should* be feasible. Pragmatically, doing so destroys everything. No, I don't even understand it myself. @beartype's support for PEP 563 (i.e., `from __future__ import annotations`) catastrophically broke, which is the ultimate harbinger of pain in this world.

To preserve future sanity, I'll *probably* drop that coercion in my next commit. It's horrifyingly fragile. I think it has something to do with caching, but... even I have no idea what is going on there. I now have no choice but to disappoint everyone by forcing everyone to explicitly pass a `BeartypeHintOverrides({...})`-style frozen dictionary rather than a `{}`-style mutable dictionary. It's for the good of everyone as well as my future psychiatrist that I cannot yet afford. I weep. :sob: 

Thanks again for all the GitHub hilarity, excitable PRs, and volunteer enthusiasm. Together, we will force this @beartype thing to finally do the RightThing™.

felixchenier:
> is what I would be saying if literally everything wasn't broken! 

OMG you got me so much it almost hurts! Kudos for actually making me believe until the end! (PS I'm glad that you don't listen to me so much, it gives me less responsibility for potential bad choices 😆)

*No seriously BeartypeHintOverrides is perfectly fine.*

> I can't help but question whether you locally ran unit tests with pytest on your end? Cause... it kinda seems like you didn't.

Ouch, now it hurts for real... I confess, I didn't run pytest. It's a shame because I always do it for my own code, with **no exception**. But I made a series of wrong assumptions that led to this bad situation: (Bad assumption 1) the test suite for beartype must be excessively complex and require every supported version of Python installed in some virtual machine; (Bad assumption 2) in this case, this is the role of GitHub's CI tools to run the unit tests so that I don't have to; (Bad assumption 3) ok all tests fail, but this is like in my first PR, it's probably only MyPy thay cries and don't let the rest process, @leycec will take care of it; (Bad assumption 4 - yours) @felixchenier knows how to code, he must have run the unit tests.

And here it is, the spaceship crashes on the moon.

Sorry for this, and a high pile of thanks for solving it. I know that I dump a great amount of work in your backyard and that because of me, you probably played a lot less video games. But I really appreciate because with these features, I can finally go on. I'm febrile that a future release will support all this, and I already started a branch of my package where I'm dumping full of boilerplate and poorly-written code in favour of one little decorator.

So happy.

PR #325, opened 2024-02-14
Add FAQ entry for `pytest-beartype`

### LINKED ISSUE #200: [Documentation] "`@beartype` + `pytest` ftw!!!" — Smart Guy @Jasha10
### Discussed in https://github.com/beartype/beartype/discussions/199

<div type='discussions-op-text'>

<sup>Originally posted by **Jasha10** December 18, 2022</sup>
The thesis here is that pytest tests and fixtures can (and should!) be decorated with `@beartype`.
Doing so makes your tests robust against mis-typed fixtures and mis-typed `@mark.parametrize` decorators.

## Motivation
Sometimes using the pytest framework is risky:

```python
from pytest import fixture

@fixture
def foo_int() -> int:
    return 123

def test_foo(foo_int: str) -> None:
    assert foo_int == "abc"
```
It's easy to see that the above test will fail since `123 != "abc"`. But can you spot the type error?
```text
$ pytest test_foo.py
...
FAILED test_foo.py::test_foo - AssertionError: assert 123 == 'abc'
1 failed in 0.05s
```
Running static checkers like `mypy` and `pyright` on the above code will *not* produce any errors.
... But the `foo_int` argument passed to the `test_foo` function actually doesn't have type `str`,
despite what the type annotation `foo_int: str` would lead you to believe.
I guess static typing can't solve everything.

## Enter beartype!

With the addition of a decorator, you can ensure that your mis-typed code will crash spectacularly:
```python
from beartype import beartype
from pytest import fixture

@fixture
def foo_int() -> int:
    return 123

@beartype  # Add decorator here!
def test_foo(foo_int: str) -> None:
    assert foo_int == "abc"
```
```text
$ pytest test_foo.py -q
...
E   beartype.roar.BeartypeCallHintParamViolation: @beartyped test_foo.test_foo() parameter foo_int=123 violates type hint <class 'str'>, as int 123 not instance of str.

<@beartype(test_foo.test_foo) at 0x7f18eac17130>:22: BeartypeCallHintParamViolation
====================== short test summary info =======================
FAILED test_foo.py::test_foo - beartype.roar.BeartypeCallHintParamViolation: @beartyped test_foo...
1 failed in 0.10s
```
The test doesn't run! We are immediately given a bear-typed slap on the wrist.
The above diagnostic info is enough to piece together that the signature of `test_foo` is mistyped.

What advantages are there to using `@beartype` with pytest? Why should I care whether my test fails with a `BeartypeCallHintParamViolation` or with a standard pytest `AssertionError` as in the first example above?
Using `@beartype`, you can follow the Single Responsibility Principle in your test code. Your test functions' body needs only to worry about test logic, not about verifying that the fixtures or `parametrize`'d arguments are correctly-typed.

## Recipes:

Beartype works with pytest tests:
```python
from beartype import beartype
@beartype  # ensure `foo_int` *actually* has type `str`
def test_foo(foo_int: str) -> None:
    assert foo_int == "abc"
```
Beartype works with pytest fixtures:
```python
from beartype import beartype
from pytest import fixture
@fixture
@beartype  # ensure the `foo_int` fixture *actually* returns an `int`
def foo_int() -> int:
    return 123
```
Beartype can make sure your `@mark.parametrize` arguments are up to snuff:
```python
from beartype import beartype
from pytest import mark
@mark.parametrize("arg", [1, 2, "three"])
@beartype  # Ensure that `arg` *actually* has type `int`
def test_small_ints(arg: int) -> None:
    assert arg < 4
```</div>

tusharsadhwani:
Ideally resolves #200

### DIFF
--- a/doc/src/_links.rst
+++ b/doc/src/_links.rst
@@ -109,6 +109,8 @@
    https://ports.macports.org/port/py-beartype
 .. _beartype PyPI:
    https://pypi.org/project/beartype
+.. _pytest-beartype PyPI:
+   https://pypi.org/project/pytest-beartype
 
 .. # ------------------( LINKS ~ beartype : package : meta   )------------------
 .. _Libraries.io:
--- a/doc/src/faq.rst
+++ b/doc/src/faq.rst
@@ -1203,6 +1203,13 @@ your GigaChad IDE of choice.
 .. #
 .. # Beartype: *because you no longer care what static type-checkers think.*
 
+*********************************************************
+How do I \*ONLY\* type-check while running my test suite?
+*********************************************************
+
+If you use pytest_, you're in luck!
+Check out the `aptly-named pytest-beartype<pytest-beartype_>`__ package.
+
 **************************************
 How do I \*NOT\* type-check something?
 **************************************
--- a/doc/src/faq.rst
+++ b/doc/src/faq.rst
@@ -1208,7 +1208,7 @@ How do I \*ONLY\* type-check while running my test suite?
 *********************************************************
 
 If you use pytest_, you're in luck!
-Check out the `aptly-named pytest-beartype<pytest-beartype_>`__ package.
+Check out the `aptly-named pytest-beartype<pytest-beartype PyPI_>`__ package.
 
 **************************************
 How do I \*NOT\* type-check something?


leycec:
**Fantastical.** Thanks so much for this pump-action resolution. It's in! Now, we party. Thursday is the new Friday. :partying_face:

PR #384, opened 2024-05-23
Update install.rst so joke is more scientifically accurate

Kevin-Mattheus-Moerman:
Replaced the word venomous with vicious, since anacondas are not venomous :)


leycec:
I see your incredible changeset and raise you Anaconda-san, the slithering Asian salary man.


<sup>*the things the eyes of Anaconda-san have seen*</sup>

Issue #448, opened 2024-09-28
[CI] Transition `tox`-based testing from `pip` 😭 to `uv` ⚡


minmax:
## [uv](https://github.com/astral-sh/uv) is super fast.

It's pretty easy to install into the system and once you try it, something switches in you, so it's hard to go back.
```bash
# On macOS and Linux.
$ curl -LsSf https://astral.sh/uv/install.sh | sh

# Linux (*deb)
snap install astral-uv

# On Windows.
$ powershell -c "irm https://astral.sh/uv/install.ps1 | iex"

# With pip.
$ pip install --user uv
```

`tox-uv` test-only dependency will be required for `uv` support in `tox`.

Additionally it may be useful to create a `make install` command that installs `uv` into the home directory and creates a `./.venv` with dependencies installed. But the user will have to edit their `~/.shrc` themselves (to modify PATH).

### And the reason:
Installing all supported pythons, creating environments, installing all dependencies and running tests took me 5 minutes 30 seconds on AMD RYZEN 5 2600 - 3.4/3.9Ghz (turbo).

@leycec pull request approval is possible for this issue?

Moosems:
It may be fast, but it is neither widely adopted nor the accepted standard as pip is. I imagine leycec would appreciate this as an alternative but simply removing pip as an option will very likely be shut down.

m3at:
Uv is cool<sup>and it's written in rust, which gives it an immediate x1000 street-cred boost!</sup>, I use it for most of my new projects and love it. But... is it really useful for beartype? Hatch can [use uv for faster dependencies resolution and env management](https://hatch.pypa.io/latest/how-to/environment/select-installer/#enabling-uv), but is speed currently a problem? If not, adding an extra tool to solve a non-problem might not be needed

leycec:
**Woops.** Extreme apologies for failing to promptly attend to this or your corresponding PR, @minmax. I promised myself I would never "be that GitHub guy." Yet, I'm being that GitHub guy. :cry: 

On the one tentacle, I undeniably like the sound of fixing this...

> Installing all supported pythons, creating environments, installing all dependencies and running tests took me 5 minutes 30 seconds on AMD RYZEN 5 2600 - 3.4/3.9Ghz (turbo).

**That's maybe horrible.** How much of that is just the "running tests" part? If most of that is just the "running tests" part, that's fine for certain definitions of fine. If most of that is *not* just the "running tests" part, that's a new breed of awful.

[...]

PR #449, opened 2024-09-28
Replace pip with uv 

minmax:
Fix for https://github.com/beartype/beartype/issues/448

CI fixes will be delivered a bit later.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,3 @@
+install:
+	[ -f ~/.cargo/bin/uv || curl -LsSf https://astral.sh/uv/install.sh | sh
+	uv sync --all-extras
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -462,7 +462,8 @@ test = [
 
     # A relatively modern version of "tox" is required. It's probably quite old
     # by now. I feel depressed just reading comments like this...
-    'tox >=3.20.1',
+    'tox >=4.20.0',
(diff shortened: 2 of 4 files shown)


leycec:
**...guh.** I feel so awful for closing your PRs without merging, @minmax. I *really* wish we could have discussed this a bit before you went to so much trouble. Sometimes, GitHub just feels bad. This is one of those times. :sob: 

Please see [my commentary here on the associated issue](https://github.com/beartype/beartype/issues/448#issuecomment-2492764439) for why this PR is (*...yet again*) being closed without merging. The **tl;dr** is that I'd strongly prefer a conditional workflow that only preferentially rather than mandatorily leverages `uv` rather than `pip`. Specifically, I'd *love* `tox` to be configured via the top-level `tox.ini` file to:

* If `uv` is installed, `tox` will install dependencies via `uv`.
* Else, `tox` will install dependencies via `pip`.

If somebody who is not me would like to make that happen, I'll immediately merge that PR without question. But I'll probably *never* merge a PR that unconditionally throws out an industry standard like `pip` in favour of a non-standard like `uv` – even if that non-standard is ostensibly faster. Speed matters... *but only so much.* Robustness and portability are significantly more useful metrics than speed for a Python package manager.

I'm so sorry, @minmax. Forgive me for these GitHub sins I have committed against you. :pray:

PR #457, opened 2024-10-01
Ruff ignore format

minmax:
Improved and supplemented version (I hope it's not boring yet)

only 3 rules:
* [Pyflakes](https://docs.astral.sh/ruff/rules/#pyflakes-f) (F) - 
* [pycodestyle-warnings](https://docs.astral.sh/ruff/rules/#warning-w) (W)
* [pyupgrade](https://docs.astral.sh/ruff/rules/#pyupgrade-up) (UP)


leycec:
**Gah!** `ruff` autofixes are proliferating faster than Tribbles on the OG NCC-1701 class USS Enterprise. Predictably, I hate most of these. Explicit is better than implicit. Thus, I strongly prefer `class MuhClass(object):` to `class MuhClass:`. The latter reads *super*-weird to me. You're always subclassing *something* – so just come out with it already. There's no tangible benefit to "hiding" the fact that you're subclassing the root `object` superclass.

The only `ruff` autofixes I mildly like here are the ones **removing extraneous trailing whitespace.** I *totally* agree with those, actually. Extraneous trailing whitespace irrationally bugs me so much. I even have a `vim` plugin that explicitly highlights extraneous trailing whitespace in red across the current buffer *and* tags all such whitespace as erroneous in the `quickfix` list. *Annnnnnnnnyway.* Boring `vim` stuff. :smile: 

If you'd like to reduce this to just the following, I'll probably merge that immediately without punching myself in the head *too* much:

* **Remove extraneous trailing whitespace.**
* **Reorder imports lexicographically** (i.e., alphabetically).

## B-b-but... Is There Any Point, Really?

Honestly, all of this is bit more work than is really desirable. Most whitespace is already clean; likewise, most imports are already ordered lexicographically. It's not like any of these things are actual errors, anyway. I disagree with `ruff` so strongly on most things that there's not much left for `ruff` to do at the end of the day.


`ruff`: *you are a bad doggo.*

JWCS:
@leycec You're not auto deleting trailing whitespace? This is the vim snippet I stole from the internet, and it hasn't let me down. On file save, strips all trailing whitespace. Never think about it again.
``` vim
" Strip trailing whitespace from file!
fun! <SID>StripTrailingWhiteSpaces()
  let l = line(".")
  let c = col(".")
  %s/\s\+$//e
  call cursor(l, c)
endfun

augroup vimrc
autocmd!
autocmd BufWritePre * :call <SID>StripTrailingWhiteSpaces()
augroup END
```
(There's also alternative ways)

PR #475, opened 2025-01-02
Fix pep 604 optional type example

ilyapoz:
Specifying `| None` is still required.
Checked that beartype actually raises an exception without it when passed `None`


leycec:
**Guh!** Thanks so much for the wondrous doco fix. You're... totally right as usual. This is a shameful error that's probably been lingering in our docos for literally decades.

Tonight, the last doco bug dies. :boot: :bug:

PR #484, opened 2025-01-21
Remove incorrect `self` argument from `mock_types` example

adamtheturtle:
Without this change, running the example gives:

```
TypeError: muh_func() missing 1 required positional argument: 'muh_arg'
```


leycec:
**Yikes.** That was serious badness. Clearly, nobody tested nuthin'. Thanks so much for this *excellent* docos resolution. Let's merge this eighth wonder of the QA world. **BAM**:exclamation:

PR #485, opened 2025-01-21
Avoid static type checking error in mock types FAQ example

adamtheturtle:
If I use the example code without this change, I get the following errors from `mypy` and `pyright`:

```
error: Read-only property cannot override read-write property  [misc]
```

```
error: "__class__" incorrectly overrides property of same name in class "object"
    Property method "fset" is missing in override (reportIncompatibleMethodOverride)
```

### DIFF
--- a/doc/src/faq.rst
+++ b/doc/src/faq.rst
@@ -808,6 +808,9 @@ type is an instance of the type it mocks, though?
    >>> class MockType:
    ...     @property
    ...     def __class__(self) -> type: return OriginalType
+   ...     @__class__.setter
+   ...     def __class__(self, value: type) -> None:
+   ...         super.__class__ = value
 
    >>> from beartype import beartype
    >>> @beartype


doc/src/faq.rst
```diff
@@ -808,6 +808,9 @@ type is an instance of the type it mocks, though?
    >>> class MockType:
    ...     @property
    ...     def __class__(self) -> type: return OriginalType
+   ...     @__class__.setter
+   ...     def __class__(self, value: type) -> None:
+   ...         super.__class__ = value
```
adamtheturtle, line 813: I'm not sure what to do here. Maybe `raise NotImplementedError`?

leycec:
**...lol.** Thanks so much for catching this really ugly edge case. You're the ultimate QA turtle. Did I mention that we love turtles where we live? Everybody does here. We live in the cold wetlands of Ontario, Canada – where mosquitoes breed by the billions. We've all got bumper stickers like "I stop for turtles." Your GitHub username warms my heart.

What were we talking about again? Oh. Right. Type mocking. Pretty bizarre stuff, huh? Static type-checkers aren't *wrong* about this, per say – but they're not necessarily *right* either. Overriding `__class__` as a `@property` is already so far off into the DeepEndOfDarkPython™ that static type-checking doesn't even really apply anymore. `mypy` and `pyright` *always* choke whenever you start overriding dunder attributes. They just can't cope with Python's penchant for dark magic, bubbling cauldrons, and ill-determined runtime semantics.

Your initial implementation of `@__class__.setter` seems fine, too. Sure... it's pretty odd. Nobody should be resetting the `__class__` dunder attribute. But if they do, at least your approach preserves working code semantics. Raising `NotImplementedError` instead would *probably* break those semantics with an unreadable exception. Maybe. Anyyyyway.

It's all baseless conjecture and pride-posturing at this point. Probably nobody cares about `@__class__.setter` except `mypy` and `pyright` – and even they don't *really* care. This is way outside their narrow wheelhouse and theological mandate from ~~Heaven~~ Guido.

Let us just ship your amazing fix. **BOOM.** It's in. This head exploding proves that you're awesome: :open_mouth: :exploding_head:  :boom:

adamtheturtle:
Thank you @leycec !

PR #511, opened 2025-03-17
Fix remaining mypy errors related to TypeForm

davidfstr:
Explanation of changes:

---

1. pep484585container.py:72: error: Incompatible types in assignment (expression has type "<typing special form>", variable has type "TypeForm[Any]")  [assignment]

> 72: hint_reduced = Annotated[Collection[Tuple[hint_key, hint_value]], ...]

mypy is having trouble with `Tuple[hint_key, hint_value]` because (1) it's not a valid type expression (because it has variables `hint_key` and `hint_value`) and (2) it's not a value expression that is of type TypeForm, because typeshed doesn't currently have definitions like:

```
class Tuple:
    def __class_getitem__(*Es: TypeForm) -> TypeForm[Tuple[*Es]]: ...

class Collection:
    def __class_getitem__(E: TypeForm) -> TypeForm[Collection[E]]: ...

class Annotated:
    def __class_getitem__(T: TypeForm, *values) -> TypeForm[Annotated[T, None]]: ...
```

Until typeshed gets those kinds of definitions, we'll need to use a `# type: ignore` for now.

---

2. redpep484585generic.py:90: error: Incompatible return value type (got "<typing special form>", expected "TypeForm[Any] | HintSanifiedData")

> 60: def resolve_hint(hint: str, ...) -> Hint:
> 90:     return hint

Cannot convert `str` directly to a TypeForm because the typechecker can't prove that the `str` spells a valid TypeForm. Fixed by changing `hint: str` to `hint: Hint`, since all callers pass in a Hint (that happens to also be a str).

---

3. redpep484585generic.py:90: error: Incompatible return value type (got "<typing special form>", expected "TypeForm[Any] | HintSanifiedData")  [return-value]

> 28: def reduce_hint_pep484585_generic_subscripted(
> 29:    hint: Hint, exception_prefix: str, **kwargs) -> HintOrHintSanifiedData:
> 90:         return Any

mypy is having trouble because `HintOrHintSanifiedData` is a `Union[Hint, other_stuff]` rather than a direct `Hint`.

Hmm. This is an interesting edge case. PEP 747 says:

> When a static type checker encounters [a type expression], the evaluated type of this expression should be assignable to `TypeForm[T]` if the type it describes is assignable to `T`.

PEP 747 does not say:

> When a static type checker encounters [a type expression], the evaluated type of this expression should be assignable to `TypeForm[T]` or `TypeForm[T] | *anything*` if the type it describes is assignable to `T`.

If PEP 747 did have the latter definition, the `return Any` would have been accepted. Unfortunately the latter definition - although intuitive - is not straightforward to implement safely in mypy... **I'll ask about this issue in the design thread for TypeForm.**

For now can workaround with (1) `return Hint(Any)` or (2) `_AnyHint: Hint = Any` + `return _AnyHint`. I'll use (2) for now since it's more performant.

---

4. redapinumpy.py:244: error: Incompatible return value type (got "<typing special form>", expected "TypeForm[Any]")  [return-value]

> 42: def reduce_hint_numpy_ndarray(...) -> Hint:
> 107: from numpy import ndarray  # pyright: ignore
> 244: return Annotated[ndarray, hint_validator]  # pyright: ignore

It looks like numpy's stubs don't declare `ndarray`, so mypy doesn't know that `ndarray` is a `Type[ndarray]`, and thus mypy doesn't know that `ndarray` can be assigned to a TypeForm. I'll add a `# type: ignore`

---

5. _convcoerce.py:191: error: Incompatible return value type (got "<typing special form>", expected "TypeForm[Any]")  [return-value]

> 67: def coerce_func_hint_root(...) -> Hint:
> 191:         return Union[hint, NotImplementedType]  # pyright: ignore

Same kind of problem as (1). Typeshed currently lacks a definition like:

```
class Union:
    def __class_getitem__(*Ts: TypeForm) -> TypeForm[Union[*Ts]]: ...
```

### DIFF
--- a/beartype/_check/convert/_convcoerce.py
+++ b/beartype/_check/convert/_convcoerce.py
@@ -188,7 +188,7 @@ def coerce_func_hint_root(
         # thus taken the surprisingly sensible course of silently ignoring this
         # edge case by effectively performing the same type expansion as
         # performed here. *applause*
-        return Union[hint, NotImplementedType]  # pyright: ignore
+        return Union[hint, NotImplementedType]  # type: ignore[return-value]  # pyright: ignore
 
     # Defer to the function-agnostic root hint coercer as a generic fallback.
     return coerce_hint_root(hint=hint, exception_prefix=exception_prefix)
--- a/beartype/_check/convert/_reduce/_nonpep/api/redapinumpy.py
+++ b/beartype/_check/convert/_reduce/_nonpep/api/redapinumpy.py
@@ -241,4 +241,4 @@ def reduce_hint_numpy_ndarray(
     hint_validator.get_repr = repr(hint)
 
     # Return this validator annotating the NumPy array type.
-    return Annotated[ndarray, hint_validator]  # pyright: ignore
+    return Annotated[ndarray, hint_validator]  # type: ignore[return-value]  # pyright: ignore
--- a/beartype/_check/convert/_reduce/_pep/pep484585/redpep484585generic.py
+++ b/beartype/_check/convert/_reduce/_pep/pep484585/redpep484585generic.py
@@ -24,6 +24,8 @@ from beartype._util.hint.pep.proposal.pep544 import (
 )
 from beartype._util.hint.pep.utilpepget import get_hint_pep_origin_or_none
 
+_AnyHint: Hint = Any
+
 # ....................{ REDUCERS                           }....................
 def reduce_hint_pep484585_generic_subscripted(
     hint: Hint, exception_prefix: str, **kwargs) -> HintOrHintSanifiedData:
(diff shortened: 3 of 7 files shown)


leycec:
**BRILLIANT!** Thank you *soooooooooooooo* much for spending your extremely valuable weekend on the intersection of `mypy` and @beartype. I both feel profound gratitude. You are the QA man.

This `typing` one-liner is also exceedingly clever:

```python
ANY: Hint = Any
```

I... never would have thought of that. Because I am a Canadian redneck, I probably would've just blindly chucked `type: ignore[cause-i-sed-so]` everywhere and called it a bad commit. Your fine-grained approach here and elsewhere is vastly preferable.

Let us merge this immediately for the good of everything that is holy and Pythonic. :partying_face: :snake: :face_holding_back_tears:

PR #548, opened 2025-09-12
Simplify Python pre-release version in workflow configuration

cclauss:


leycec:
Thanks so much for the PR, Big Boss @cclauss! `python-version: "3.14"` totally seems like a useful improvement, especially as we're now less than a month away from CPython 3.14's official release.

Let's merge this. :partying_face:

PR #586, opened 2025-11-13
Use 'writer' arg in publish_parts to avoid PendingDeprecationWarning

eamanu:
Replace the deprecated args in test_docreadme.py test. This prevents a PendingDeprecationWarning from docutils (which pytest was treating as an error) and keeps the test compatible with Docutils current API.

Originally reported https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1120246

### DIFF
--- a/beartype_test/a90_func/a80_doc/test_docreadme.py
+++ b/beartype_test/a90_func/a80_doc/test_docreadme.py
@@ -95,7 +95,7 @@ def test_doc_readme(monkeypatch) -> None:
 
     # Attempt to render this "README.rst" file as reST, implicitly invoking this
     # monkey-patch.
-    publish_parts(source=README_CONTENTS, writer_name='html4css1')
+    publish_parts(source=README_CONTENTS, writer='html4css1')
 
     # Assert "docutils" to have emitted *NO* warnings or errors.
     assert not system_messages


leycec:
**Superbness!** Thanks so much for the timely fix, your stupendous hair, <sup>*luv it*</sup> and making Debian what it is. Sadly, tests fail with explosions in the sky:

```python
FAILED ../../../beartype_test/a90_func/a80_doc/test_docreadme.py::test_doc_readme - 
AssertionError: passed string "html4css1" as "writer" parameter; pass an instance, or use the
"writer_name" parameter instead (in docutils.core.publish_* convenience functions).
```

Thus, I avoid merging. Since this integration test has become problematic *and* since @beartype is migrating away from the Sphinx + `docutils` stack towards Mkdocs as quickly as it can, I'll immediately push a new commit skipping this test entirely for the moment. Clearly, that's an awful "solution." I have no defense. The issue tracker is a beast with a billion backs. In 2025, forcing failing tests is just how @beartype rolls. :joy: 

Lemme know if you'd like me to cut a new @beartype `0.22.6` patch release to satisfy Debian constraints. Until then, thanks so much for the heads-up. You're awesome! I envy that hair. I really do...

leycec:
**...lol.** Nevermind. I'm dumb. It's well-known. This PR isn't about @beartype's geriatric documentation. It's just about our root `README.rst` file. Gotcha! That's a relief. Looks like we just need to require a modern `docutils >=0.22.0` version both:

* Here for this specific test.
* As a new optional dependency listed in the `test-tox` extra of our top-level `pyproject.toml` file. The issue is that almost certainly conflicts with optional dependencies already listed in the `doc-rtd` extra of that file. Oh, well. Watcha goin do. :face_exhaling: 

Let me give some of this a go. *Brb...*

leycec:
**Merged by 43856587c7b5aee.** Thanks so much for the superb fix! I've included your most excellent change in a *much* larger commit resolving both this and a medley of other pending issues. Have a great Winter! Here in Canada, we're already hip-deep in meters of snow. Truly, this is the best time of year. Let's go skiing, everybody. :ski: :skier: :skull:

Issue #603, opened 2025-12-14
[Bug] Generics subscripted by self-referential forward references to nested classes = un🐻able


danielkovtun:
**Summary**

Forward references to nested classes (e.g., 'Outer.Inner') fail at runtime when used inside generic containers like list['Outer.Inner'], but work correctly when used as bare type hints. The bear can handle a bare 'Outer.Inner', but can't bear it inside a `list[]`.

This affects both direct `@beartype` decoration and usage via `jaxtyping.install_import_hook()`.

Environment
- Python 3.11.12
- beartype 0.22.9
- jaxtyping 0.3.3 (for secondary test case)

**Reproducer**

`outer_module.py` (for direct `@beartype` tests)

```python
"""Module with nested class pattern."""
from dataclasses import dataclass
from beartype import beartype


@dataclass
class Outer:
    @dataclass
    class Inner:
        value: int

    def get_bare(self) -> 'Outer.Inner':
        return Outer.Inner(value=1)

    def get_list(self) -> list['Outer.Inner']:
        return [Outer.Inner(value=2), Outer.Inner(value=3)]


@beartype
def accepts_bare(x: 'Outer.Inner') -> 'Outer.Inner':
    return x


@beartype
def accepts_list(x: list['Outer.Inner']) -> list['Outer.Inner']:
    return x
```

`test_beartype_only.py`

```python
"""
Test: Direct @beartype with nested class forward refs in generics.
Only tests the direct @beartype case (no jaxtyping).

Run: python test_beartype_only.py
"""
import os
import sys

sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))

from outer_module import Outer, accepts_bare, accepts_list


def run(name, fn):
    try:
        fn()
        print(f"  PASS: {name}")
        return True
    except Exception as e:
        print(f"  FAIL: {name}")
        print(f"        {type(e).__name__}: {str(e)}")
        return False


if __name__ == '__main__':
    print("Direct @beartype - nested class forward refs in generics")
    print("=" * 55)
    results = []

    results.append(run("'Outer.Inner' (bare)", lambda: accepts_bare(Outer.Inner(1))))
    results.append(run("list['Outer.Inner'] (generic)", lambda: accepts_list([Outer.Inner(1)])))

    print(f"\nResults: {sum(results)}/{len(results)} passed")
    sys.exit(0 if all(results) else 1)

```

`test_jaxtyping_only.py`

```python
"""
[...]

PR #604, opened 2025-12-14
Initial attempt at supporting forward references to nested classes in generic containers

danielkovtun:
Fixes forward references to nested classes like `'Outer.Inner'` when used inside generic containers (e.g., `list['Outer.Inner']`).

## Problem

When beartype encounters a forward reference like `'Outer.Inner'`, it parses the dotted name and assumes `'Outer'` is a module. But when `Inner` is a class nested inside class `Outer`, this assumption fails and resolution breaks. 

See #603 

## Workaround

Two changes:

1. `fwdrefmake.py` - Detect nested class refs via frame inspection

When the inferred "module" name isn't importable, inspect the call stack to find the actual module:
- Look for a frame executing the class body (`f_code.co_name == base_class_name`)
- Fall back to checking if any frame's module already defines the base class

2. `utilmodimport.py` - traverse dotted attribute paths

When resolving `'Outer.Inner'` against a module, traverse the full attribute path (`module.Outer.Inner`) instead of treating 'Outer' as a module name.

NOTE: This fix adds O(n) stack traversal for the specific edge case of nested class forward references where the inferred “module” name is not importable. This cost is:
 - Only incurred when `import_module_or_none(type_module_name)` returns None
 - Memoized via the existing `_forwardref_args_to_forwardref` cache

### DIFF
--- a/beartype_test/a00_unit/a60_check/a80_forward/reference/test_fwdrefmake.py
+++ b/beartype_test/a00_unit/a60_check/a80_forward/reference/test_fwdrefmake.py
@@ -16,6 +16,14 @@ This submodule unit tests the
 # package-specific submodules at module scope.
 #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
+# NOTE: This module-scope import is required for the nested class forward
+# reference test below (see "PASS ~ nested" section). Frame inspection in
+# _make_forwardref_subtype() uses hasattr(module, 'Class') to find modules that
+# define a class. Function-local imports do NOT make the class a module
+# attribute, so we must import at module scope for the test to work.
+# See: https://github.com/beartype/beartype/issues/603
+from beartype_test.a00_unit.data.data_type import Class
+
 # ....................{ TESTS                              }....................
 def test_make_forwardref_indexable_subtype() -> None:
     '''
@@ -24,11 +32,14 @@ def test_make_forwardref_indexable_subtype() -> None:
     factory.
     '''
 
-    # ....................{ LOCALS                         }....................
+    # ....................{ IMPORTS                        }....................
     # Defer test-specific imports.
     from beartype.roar import (
         BeartypeCallHintForwardRefException,
         BeartypeDecorHintForwardRefException,
+    )
+    from beartype._check.forward.reference.fwdrefmake import (
+        make_forwardref_indexable_subtype,
     )
     from beartype_test.a00_unit.data.check.forward.data_fwdref import (
         CLASS_BASENAME,
@@ -115,6 +126,40 @@ def test_make_forwardref_indexable_subtype() -> None:
     # Assert that this representation contains the expected substrings.
     for fwdref_repr_substr in FORWARDREF_REPR_SUBSTRS:
         assert fwdref_repr_substr in FORWARDREF_REPR
+
+    # ....................{ PASS ~ nested                  }....................
+    # Test nested class forward references where the first component is a class
+    # rather than a module. This exercises the frame inspection fallback in
+    # _make_forwardref_subtype.
+    #
+    # To trigger frame inspection, we pass a scope_name that CANNOT resolve
+    # "Class". The frame inspection code will iterate through the call stack
+    # and find this test module, which has "Class" imported above. It will then
+    # use this test module as the scope for resolving the nested class.
+
+    # A scope that does NOT define "Class", forcing frame inspection fallback.
+    UNRELATED_SCOPE = 'beartype._util.module.utilmodimport'
+
+    # Create a forward reference proxy to a nested class. Since UNRELATED_SCOPE
+    # cannot resolve "Class", the code must use frame inspection to find that
+    # this test module has "Class" imported.
+    forwardref_nested = make_forwardref_indexable_subtype(
+        scope_name=UNRELATED_SCOPE,
+        hint_name='Class.NestedClass',
+    )
+
+    # Assert the forward reference proxy has the expected attributes.
+    # The scope_name should have been updated to this test module's name
+    # by the frame inspection logic.
+    assert forwardref_nested.__name__ == 'NestedClass'
+    assert forwardref_nested.__name_beartype__ == 'Class.NestedClass'
+
+    # Assert the forward reference proxy resolves to the correct type.
+    assert forwardref_nested.__type_beartype__ is Class.NestedClass
+
+    # Assert isinstance checks work with the nested class forward reference.
+    nested_instance = Class.NestedClass()
+    assert isinstance(nested_instance, forwardref_nested)
 
     # ....................{ FAIL                           }....................
     # Assert that attempting to access an undefined dunder attribute of a
--- a/beartype_test/a00_unit/a70_decor/a20_code/a60_pep/pep484/test_pep484ref.py
+++ b/beartype_test/a00_unit/a70_decor/a20_code/a60_pep/pep484/test_pep484ref.py
@@ -32,12 +32,14 @@ def test_pep484_ref_data() -> None:
     from beartype.roar import BeartypeCallHintReturnViolation
     from beartype_test.a00_unit.data.hint.data_hintref import (
         BeforeTheHurricane,
+        ClassWithNested,
         TheDarkestEveningOfTheYear,
         WithSluggishSurge,
         a_little_shallop,
         but_i_have_promises,
         its_fields_of_snow,
         of_easy_wind,
+        rugged_and_dark_winding,
         stopping_by_woods_on,
         the_dry_leaf,
         the_woods_are_lovely,
@@ -70,6 +72,12 @@ def test_pep484_ref_data() -> None:
         TheDarkestEveningOfTheYear)
     assert RUGGED_AND_DARK.or_where_the_secret_caves() is RUGGED_AND_DARK
     assert winding_among_the_springs(RUGGED_AND_DARK) is RUGGED_AND_DARK
+
+    # Test nested class forward references inside generic containers (e.g.,
+    # List['Outer.Inner']). This exercises the nested class resolution logic
+    # added to fix issue #603.
+    NESTED_LIST = [ClassWithNested.NestedClass(), ClassWithNested.NestedClass()]
+    assert rugged_and_dark_winding(NESTED_LIST) is NESTED_LIST
 
     # ..................{ FAIL                               }..................
     # Assert that calling a method violating its return annotated as a 2-tuple


leycec:
**Beautiful!** Wonderful! And other superlatives. At this critical juncture, I must publicly admit that I am a negligent maintainer. I'm about to play video games. Any ignominious issue like this that digs deep into forward references demands my full attention. Apparently, it is difficult to play video games while reviewing GitHub changes.

In lieu of a full-bore code review tonight, I do superficially note that:

* **This branch is behind `main`.** For everyone's safety, would you mind locally updating your beautiful and wonderful PR with our most recent changes from `main`? I'm so sorry. :bow: 
* **[Tests are currently failing](https://github.com/beartype/beartype/actions/runs/20219630962/job/58038905664?pr=604)**. Fortunately, it looks like a trivial static type-checking failure. Unfortunately, the failure of this one test has cascaded to cause the entire test suite to fail with this PR. Would you mind resolving this on your end? Just so that we can see if the rest of the tests are passing? This is why `mypy` hates @beartype today, apparently:

```
beartype/_util/module/utilmodimport.py:454: error: Incompatible types in assignment
(expression has type Module, variable has type "type | Iota")  [assignment]
Found 1 error in 1 file (checked 420 source files)
```

On the bright side, I *love* that you've even done a super-smart Big-Oh analysis:

> NOTE: This fix adds `O(n)` stack traversal for the specific edge case of nested class forward references where the inferred `module` name is not importable. This cost is:

> * Only incurred when `import_module_or_none(type_module_name)` returns `None`
> * Memoized via the existing `_forwardref_args_to_forwardref` cache

More beautiful. More wonderful. `O(n)` isn't great, of course. But everyone expects forward reference resolution to be expensive. You can't really get around that. Except, thanks to memoization, we totally do get around that! Due to caching, this should actually just reduce to `O(1)` amortized time complexity. Maybe. Let's pretend that's true.

I'll dig my bear paws into this tomorrow. Until then, have a great Sunday evening. Thanks so much for helping out a type-checking bear brother in need, @danielkovtun. :bear:

danielkovtun:
@leycec thanks for the kind words, happy to make a contribution to this admittedly rare, and probably convoluted use-case! But who doesn't love a healthy dose of nesting?!
image
 
I've pulled latest main and claude, that mad lad, chose to debug my original (incorrect) re-assignment of `module_attr` by shamelessly committing an ignore-assignment "fix" 😪 

Alternatively, I could avoid the re-assignment by adjusting the variable name returned in 
https://github.com/beartype/beartype/blob/121dc9304023e3112978a8fd624080f00a90cfc0/beartype/_util/module/utilmodimport.py#L398

Let me know if the ignore-assignment is a forbidden fruit in the land of the 🐻 , happy to try a different approach. 


Tests are passing for me locally. I'm also happy to add a test case for the newly introduced codepath if you'd like 👍

Glinte:
It seems to me that this is another mypy limitation where it infers the type from first usage, pyright works fine here. I believe this is a bit cleaner though, and shows the intended type of `module_attr`, but honestly a type: ignore is good enough here.

```diff
diff --git a/beartype/_util/module/utilmodimport.py b/beartype/_util/module/utilmodimport.py
--- a/beartype/_util/module/utilmodimport.py	(revision 820dbd5188ce5c9dbd47c6d14cb3963ab1008a1e)
+++ b/beartype/_util/module/utilmodimport.py	(date 1765806087399)
@@ -23,7 +23,7 @@
 from beartype._data.typing.datatyping import TypeException
 from beartype._util.error.utilerrwarn import issue_warning
 from beartype._util.text.utiltextidentifier import die_unless_identifier
-from beartype._data.kind.datakindiota import SENTINEL
+from beartype._data.kind.datakindiota import SENTINEL, Iota
 from importlib import import_module as importlib_import_module
 
 # ....................{ IMPORTERS                          }....................
@@ -389,6 +389,7 @@
     attr_basename = attr_name
 
     # If this attribute name is an unqualified basename...
+    module_attr: Any | Iota
     if is_attr_name_unqualified:
         # If either no module name was passed *OR* only an empty module name was
         # passed...
@@ -451,7 +452,7 @@
     # simple single-level attribute lookup.
     if '.' in attr_basename:
         # Iteratively resolve each component of the dotted path.
-        module_attr = module  # type: ignore[assignment]
+        module_attr = module
         for attr_part in attr_basename.split('.'):
             module_attr = getattr(module_attr, attr_part, SENTINEL)
```

leycec:
**Tests pass!** Rejoice, @danielkovtun. :partying_face: 

Thanks a heap for taking a look at this, @Glinte. I promised to review this tonight. Unsurprisingly, I failed. I blew all my open-source volunteer time tonight [setting up real-time Discord-like chat for the @beartype fandom over at `beartype.zulipchat.com`](https://beartype.zulipchat.com). Lament, @danielkovtun. :sob:

You may now be wondering...

## What Is "zulipchat"? What Does That Have to Do with This PR!?

I'll conveniently ignore the second question.

Zulip is like Discord – only with a janky UX, open-source, and free-as-in-both-beer-and-speech. Took *forever* to setup, too. No idea if I've even configured that monster properly. Completely unrelatedly, would either of you two care to bang your heads against that URL? Let me know if you can even join this thing, if you can see any channels like `#beartype`, and whether anything makes sense. It kinda makes no sense to me. But I'm a bald man with a pot belly. Many things no longer make sense to me.

Please test-drive this IRC-like monstrosity with me before it destroys Tokyo for the last time:

https://beartype.zulipchat.com


<sup>*me right now*</sup>

Glinte:
Zulip feels sucky, I would rather talk in github discussions tbh. If you're thinking about community building, it is probably a bad idea too, discord seems to be THE choice here. If you care about data ownership and privacy, then it is okay I guess, I would probably just have used [Element](https://element.io) though.

leycec:
> Zulip feels sucky...

It absolutely does. That said, I'm down with sucky UX. I misspent my youth first on Bulletin Board Systems (BBSs) and then on their less colorful but ultimately *much* more useful predecessor, Internet Relay Chat (IRC). Today, I voluntarily use ViM. Like, *original* ViM. Not even Neovim, people. Oldschool friggin' ASCII ViM. I adore decrepit 90's-era websites now on their last legs with venerable domain names like `craigslist.org`, `gamefaqs.com`, and `rateyourmusic.com`. I'm 100% Team Sucky UX. :joy: 

More importantly, most of the heavyweights in the Rust and Python community have switched to Zulip. Like:

* On the Rust side, this includes the entirety of the core Rust and `cargo` teams – [which previously used Discord](https://zulip.com/case-studies/rust).
* On the Python side, this includes `conda-forge` – [which previously used Elements/Matrix](https://conda-forge.org/news/2024/11/04/moving-to-zulip).

Those migrations happened for a reason. One does not simply abandon one's entire online community. I trust Rust and `conda-forge`. How could I not? Both have proven themselves time and time again. Julia also opened up a Zulip recently. Just another data point, but the sample size is large.

Discord and Elements/Matrix were walked away from reasons. Right? And the reasons are pretty obvious. Discord may have market share today, but... so what? Discord is gaming-centric. It's a poor fit for open-source community building. The intersection of volunteerism and the profit motive is the empty set.

And then there's Elements/Matrix, which is at the opposite end of the spectrum from Discord. It's both open and federated! On paper, that's even better than Zulip! In practice, though, Elements/Matrix lacks what makes Zulip so attractive for larger-scale projects: **organization.** Elements/Matrix is completely unorganized. It's basically web IRC. Like in IRC, you've got channels. And... that's literally it. Channels. There's no threaded conversations or folders. This is unsurprising, because Elements/Matrix has IRC bridges. It's like they intentionally designed Elements/Matrix to just be IRC except for `https://`.

Meanwhile, Zulip. It's a beast. It's an ugly beast. But it actually lets us organize on the basis of Python packages and within each Python package on the basis of topics. We have a `#beartype` channel, a `#pytest-beartype` channel, an `#ipython-beartype` channel, and a `#plum` channel – and within each of those, individuated topics. Zulip organization is unmatched.

It's also as ugly as Geocities. Because the eyes bleed when you look at Zulip, nobody may use it. If nobody uses it, that's fine. Fine, I say! The wisdom of the unruly mob has spoken. When has democracy ever let us down!?!? <sup>*oh gods oh gods*</sup> :face_exhaling:

leycec:
**Back to this issue!** Hah, hah... I *so* wanted to merge this tonight, @danielkovtun. But this isn't right at all:

```python
def _make_forwardref_subtype(
    scope_name: Optional[str],
    hint_name: str,
    type_bases: TupleTypes,
) -> BeartypeForwardRef:
    ...

    if not type_module_name:
        type_module_name = scope_name
    #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
    # OH. GODS. NO. NO!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
    #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
    elif import_module_or_none(type_module_name) is None:
```

The `import_module_or_none(type_module_name)` is the problem. That's... a dangerous function we are calling at a *very* early time to be calling such a dangerous function. Let's get specific.

The `_make_forwardref_subtype()` forward reference factory is called at `@beartype` decoration time, which is almost always called from module (i.e., global) scope, which is thus performed at **module importation time.** @beartype makes forward reference proxies when a module is imported. But what happens when two modules circularly refer to one another: e.g.,

```python
# In module "mr_t":
from beartype import beartype

@beartype
def i_pity(da_foo: 'mr_s.Robert_Rocky_Balboa')

class B_A_Baracus: ...
```

```python
# In module "mr_s":
from beartype import beartype

@beartype
def yo_adrian(its_me: 'mr_t.B_A_Baracus')

class Robert_Rocky_Balboa: ...
```

So. What happens now? Let's assume the module `mr_t` is imported from *before* the module `mr_s` is imported from. Then @beartype sees the forward reference `'mr_s.Robert_Rocky_Balboa'` and coerces that into a proxy by calling `_make_forwardref_subtype()`. If we accept this changeset, that then invokes `import_module_or_none('mr_s')`. So far so good. But what happens now?

`import_module_or_none('mr_s')` dynamically imports the module `mr_s`. I trust we all see what is about to happen. While `mr_t` is being imported, `mr_s` is imported. @beartype sees the forward reference `'mr_t.B_A_Baracus'` and coerces that into a proxy by calling `_make_forwardref_subtype()`. If we accept this changeset, that then invokes `import_module_or_none('mr_t')`.

This is no longer good. In fact, this is **extreme badness.** the module `mr_t` has yet to be fully imported and is thus only partially initialized. [Importing modules that are only partially initialized is a *huge* "Nope! Nope!" in Python](https://rollbar.com/blog/how-to-fix-circular-import-in-python/#). We have a simple term to describe this situation: **circular imports.** But... *wait a minute.*

Forward references exist exactly to circumvent the issue of circular imports in the first place! If @beartype starts performing circular imports at decoration time to disambiguate between different kinds of forward references, then @beartype is **destroying those same forward references it is supposed to be saving.**

This reminds me of that [infamous Vietnam-era quote from an unknown US Army Major](https://www.thisdayinquotes.com/2015/02/it-became-necessary-to-destroy-the-town-to-save-it), who wisely still remains anonymous to this very day:

> "It became necessary to destroy the town to save it."

@beartype is not destroying the app to save the app. Let's find a better way.

leycec:
**Redux. Rewind. Resurrect.** Let's back up. I'm pretty sure this PR can be salvaged. In fact, I know it can! I believe in your formidable prowess, @danielkovtun. You posted a cat webdev meme. Truly, you are like me. And people like us? People who post cat webdev memes that only the elect 1% of 1% of humanity will ever understand, let alone crack a toothy grin at? *We. Never. Give. Up.*

Actually, you might just wanna give up right now. You can soldier on, but it means going even deeper into the magical innards of the @beartype codebase. Here's what one or both of us needs to do now:

* Shift out the *entirety* of that new problematic `elif import_module_or_none(type_module_name) is None:` branch into **the existing `beartype._check.forward.reference.fwdrefmeta.BeartypeForwardRefMeta.__type_beartype__()` property.**

That's it. But... that's still a lot. It may not sound like much, but the `BeartypeForwardRefMeta` metaclass is a *very* smelly and different beast altogether. It smells. I can smell the stench from here. It smells like... *victory.* Alright! Alright already. I'll stop with the super-lame and increasingly awkward Vietnam flic one-liners.

First, take a look at the body of that `__type_beartype__()` property. See the existing call to `import_module_attr()`? Right. We'll need to refactor that into... *something else entirely.* For both simplicity and efficiency, we could just call `import_module_attr_or_none()` instead. That ensures that the common case continues to be handled efficiently.

Once we do that, we check whether the `import_module_attr_or_none()` return value is `None` or not. If it's `None`, we do what your `elif import_module_or_none(type_module_name) is None:` is currently doing. In the end, this is what I'm envisioning the resulting code to resemble:

```python
    @property
    def __type_beartype__(cls: BeartypeForwardRef) -> type:  # type: ignore[misc]
        ...

        # Referent referred to by this forward reference proxy if a prior access
        # of this property has already resolved this referent *OR* "None" (i.e.,
        # if this is the first access of this property).
        referent = _forwardref_to_referent_get(cls)

        # If this referent has yet to be resolved, this is the first call to
        # this property. In this case...
        if referent is None:  # type: ignore[has-type]
            # print(f'Importing forward ref "{cls.__name_beartype__}" from module "{cls.__scope_name_beartype__}"...')

            # Exception subclass and prefix to be raised below.
            EXCEPTION_CLS = BeartypeCallHintForwardRefException
            EXCEPTION_PREFIX = 'Forward reference '

            # Forward referent dynamically imported from this module.
            referent = import_module_attr_or_none(
                attr_name=cls.__name_beartype__,
                module_name=cls.__scope_name_beartype__,
                exception_cls=EXCEPTION_CLS,
                exception_prefix=EXCEPTION_PREFIX,
            )

            #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
            # INSERT YOUR SPECIAL MAGICAL DARKNESS HERE
            #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
            if referent is None:
                # Unqualified basename of the outermost class in this nested reference
                # (e.g., "Outer" for the forward reference "Outer.Inner").
                base_class_name = hint_name.partition('.')[0]

                # True only if "scope_name" refers to a module defining "base_class_name".
                scope_can_resolve = False

                # If a scope name was passed, check if that scope can resolve this
                # nested class reference.
                if scope_name is not None:
                    ...
```

Right? Something like that. I'm waving my noodly appendages like a flying spaghetti monster here. :spaghetti: 

You'll note that the above code is getting *really* indented, though. It'd probably be reasonable to shift out your nested class-specific logic into a new private `_import_nested_class()` function or something.

## That Said, I'm Getting an Ugly Feeling Here

Honestly, I'm unconvinced any of this works as advertised in the general case. Consider an arbitrary absolute forward reference `"a.b.c.d.e"`. Literally *any* of `a`, `b`, `c`, `d`, or `e` can be a class. Moreover, *any* of `a`, `b`, `c`, or `d` (but *not* `e`) can also be a (sub)package or (sub)module. Whichever of the first of those is a class then partitions what precedes that to be (sub)packages and (sub)modules *and* what follows that to all be nested classes.

We have to decide that partitioning. This PR... isn't doing that. This PR blithely assumes that *all* forward references are relative. They aren't! Many forward references are absolute.

Consider a forward reference `"Module.Outer.Inner"`, for example. *We* know that `Module` is a module name, `"Outer"` is a globally scoped class name in that module, and `"Inner"` is a nested class name in the former class. But how does *@beartype* mechanically decide and know those things?

In other words, it's a combinatorial problem. And the time complexity seems *way* worse than merely `O(n)`. Doesn't it? I'm pretty sure it's `O(nk)` now for `n` the current stack height and `k` the number of `"."` delimiters in the current forward reference. But `n == k` in the worst case! This has thus gone quadratic to `O(n**2)`. That's... not good.

Maybe that's what @beartype needs to do to distinguish this worst case. But maybe @beartype can also just advise everyone to avoid using forward references to nested classes in the first place with a FAQ entry. Like, users could just make aliases to nested classes ala:

```python
"""Module with nested class pattern."""
from dataclasses import dataclass
from beartype import beartype


@dataclass
class Outer:
    @dataclass
    class Inner:
        value: int

    def get_bare(self) -> 'Outer.Inner':
        return Outer.Inner(value=1)

    def get_list(self) -> list['Outer.Inner']:
        return [Outer.Inner(value=2), Outer.Inner(value=3)]

# Alias to inner class, yo! Super lame. But what you gonna do?
Outer_Inner = Outer.Inner

@beartype
def accepts_bare(x: 'Outer_Inner') -> 'Outer_Inner':  # <-- heh
    return x


@beartype
def accepts_list(x: list['Outer_Inner']) -> list['Outer_Inner']:  # <-- heh heh
    return x
```

*...yeah.* It's super-dumb. I know. But it also works *right now* without touching fragile @beartype internals and preserves @beartype efficiency. Those things have value.

## Let Us Now Beg a Higher Power to Solve This for Us

I also acknowledge that @beartype should definitely resolve this issue properly at some point. But how do we do that *without* invoking worst-case quadratic complexity? Hmmmmmmmmm...

I bet we could design some sort of a clever local cache like so:

1. While we walk up the call stack, we could maintain a local dictionary `frame_name_to_obj: dict[str, FrameType] = {}` mapping from the unqualified basename of each module or class object encapsulated by the currently iterated frame to that module or class. Hmm..... No idea. Might be overkill. Anyway! *Onwards.*
2. We iterate up the call stack until we visit a frame such that either:
   * This frame encapsulates a type. If the unqualified basename of this frame is that of the "next" nested forward reference in `hint_name.split('.')`, then this type is a possibly nested local class and **this reference is possibly relative**. Else, it isn't and **this reference is definitely absolute.** 
   * A `<module>` boundary. If this happens, then we know **this reference is definitely absolute.**
3. We continue this iteration until we visit a frame that also encapsulates a type *and* the unqualified basename of this frame is that of the **root referent** `hint_name.partition('.')[0]`. In this case, this type is *definitely* a nested local class and **this reference is definitely relative.**

The whole crux of the algorithm is differentiating what `"."` delimiters mean. Are we dealing with nested classes in a relative forward reference (e.g., `"Outer.Inner"`) or are we dealing with nested modules in an absolute forward reference (e.g., `"package.module.submodule.Type"`)?

Ambiguity in the `"."` delimiter is the obvious problem. Dealing with that ambiguity requires a non-obvious but "fun" solution. Let someone else deal with this Hell! Please, GitHub! I beg only this one thing of you! :pray:

## Simplify the Brutality before Someone's Head Explodes

Clearly, this is a Big Problem™. Thankfully, it's also a Big Problem™ almost no one cares about. As with all big problems in Computer Science, we solve them by subdividing them into smaller problems.

In this case, it seems like it would probably be best to **avoid trying to solve the general case just yet.** Initially, let's just work on solving the exact case of a relative forward reference of the form `"Type.NestedType"`. We won't even worry about absolute forward references containing nested classes (e.g., `"package.module.Type.NestedType"`) or relative forward references containing two or more nested classes (e.g., `"Type.NestedType.NestedNestedType"`) just yet.

`"Type.NestedType"` is probably the most common idiom, anyway. It's doubtful anyone cares about nested *nested* types. Of course, even nailing this exact use case will probably prove non-trivial. But at least it makes things a little saner and more tractable.

My kingdom for a cat webdev meme.


<sup>*holy balls it's so true*</sup>

leycec:
**WAIT.** We're *totally* barking up the wrong tree here. None of this should be resolved with **forward reference proxies** (i.e., the `beartype._check.forward.reference` subpackage). Why? Because @beartype already internally passes around the full **nested class stack** via the `cls_stack: tuple[types, ...] | None = None` parameter.

This specific issue is interested only in resolving relative forward references to the nested class currently being declared, right? That's `cls_stack`. No call stack frame introspection should be required. At all. Because @beartype already knows and internally maintains a list of *all* nested classes currently being declared, this should actually be pretty trivial *and* efficient to solve. No cray-cray `O(n)` or `O(nk)` algorithms are needed here. Phew! :relieved: 

I... should have seen that earlier. There is so much magic going on in the @beartype codebase that it blinds even me.

## Let Me Tell My Older Self How to Solve This

Okay, older self. Let's do this. This isn't too hard, honestly. We just have to lightly refactor our higher-level `make_check_expr()` dynamic code generator to properly pass the `cls_stack` parameter to the lower-level `express_func_scope_type_ref()` function. <sup>*oh god oh god*</sup>

That is to say, we want the `express_func_scope_type_ref()` signature to resemble:

```python
def express_func_scope_type_ref(
    # Mandatory parameters.
    func_scope: LexicalScope,
    forwardref: Pep484585ForwardRef,
    refs_type_basename: Optional[set],

    # Optional parameters.
    cls_stack: TypeStack = None,  # <-- *THIS IS NEW MAGIC*
    exception_prefix: str = 'Globally or locally scoped forward reference ',
) -> Tuple[str, Optional[set]]:
```

Easy. Then, we need to refactor *all* calls to `express_func_scope_type_ref()` in `make_check_expr()` to pass that parameter: e.g.,

```python
                (
                    hints_meta.hint_curr_expr,
                    hint_refs_type_basename,
                ) = express_func_scope_type_ref(
                    forwardref=hint_curr,  # type: ignore[arg-type]
                    refs_type_basename=hint_refs_type_basename,
                    func_scope=hints_meta.func_wrapper_scope,
                    cls_stack=cls_stack,  # <-- *NEW MAGIC HERE*
                    exception_prefix=EXCEPTION_PREFIX,
                )
```

Easy-peasy. The final step is not *quite* as easy-peasy, but should still be pretty easy-peasy. We need to internally refactor the `express_func_scope_type_ref()` function to actually introspect the passed `cls_stack` parameter. Specifically, this assumption is overly simplistic and thus bad:

```python
    # If either...
    if (
        # This reference was instantiated with a module name...
        ref_module_name or
        # This classname contains one or more "." characters and is thus already
        # (...hopefully) fully-qualified...
        '.' in ref_name
    # Then this classname is either absolute *OR* relative to some module. In
    # either case, the class referred to by this reference can now be
    # dynamically imported at a later time. In this case...
    ):
```

Rather than simply assuming that `'.' in ref_name` means "absolute forward reference," we'll want to additionally test `ref_name` against the passed `cls_stack`. Not too hard. We leave this as an exercise to the reader. For tonight...

We sleep like the dead! Literally. I feel like I'm dead, guys.


<sup>*@leycec if he still had hair*</sup>

leycec:
**...lol.** It's beginning to slowly dawn on me that @beartype is so magical, so dark, and so depraved it literally cannot be maintained by anyone except me. And even I can barely maintain @beartype. I'm pretty sure this is not a healthy state of affairs. Yet, it's what we've got.

I've assiduously documented *everything* in @beartype a gazillion times over. Still, it's not enough. It's never enough! Dynamic code generation is just not going to get any easier with time, is it? In other words...

## Issue #603 Annoys Me to Heck. To Heck... and Back!

I am so annoyed I now have no choice but to fix this myself. That's fine, though. That's fine. I keep saying that. Yet my face is exhaling. :face_exhaling: 

Your first-draft attempt was a monumental inspiration, @danielkovtun. Thank you so much for motivating me to resolve this. Left to my own devices, I'd have just ignored issue #603. Nested classes are fun – but nobody uses them. Gotta issue triage. Right?

**Wrong.** To heck with issue triage. Nobody may use nested classes... but real-world practicalities no longer matter. Issue #603: "Your days lurking on this issue tracker are numbered."

This PR is demonstrably busted. My hollow promises to resolve this truly leave this PR with nowhere to go. Let's quietly close this in favour of me doing something rather than typing endless paragraphs into the gaping void of GitHub like I usually do. :joy:

danielkovtun:
Hey @leycec, apologies for the delayed response, been a busy week and wanted to fully digest your replies! 📚 

Thanks for the deep dive into why the `import_module_or_none()` approach was fundamentally flawed. The circular import issue you outlined makes total sense in hindsight, forward references exist to avoid imports at decoration time, not trigger more of them. 

The `cls_stack` approach is way cleaner. If you want a second pair of eyes when you get to it (or someone to write tests), happy to help. Otherwise, I'll keep an eye on the issue and if nothing else, I learned a lot from this detour into beartype's dark magical innards (even if I emerged slightly singed).

Thanks for the patience and the memes 😹

leycec:
**Merry Pythooooooonmas!** The happiest bug-crunching season of the year is yet again upon us. Thanks for being so understanding about all this. I deeply regret rejecting your otherwise stirling PR, but... I didn't know what else to do. I myself still hardly know how to resolve this. It's shed a light into dark corners of the codebase I've long hoped would just die already.

Hoo boy. :sob:

> If you want a second pair of eyes when you get to it

**Hah, hah!** I'm actually getting to it as we speak. Commit b7e158af630d75e is now the seventh of a growing commit chain trying to address this before [the year 2077 – apparently also the year a cybernetic ghost Keanu Reeves AI jacks straight into our heads, not that I'm opposed per say](https://www.youtube.com/watch?v=IoRWhAAYPhk). What happened to Christmas...

Instead of crunching through my frightful backlog of a bazillion video games, I've been doing this. Why!? It all started out so fun, fanciful, and carefree. But somewhere along the way, I realized I was refactoring @beartype's entire forward reference backend. The Road to QA Perdition is dimly lit and populated only by gibbering shadows.

Come check out #603 for an endless litany of commits that will sear the back of your eyelids. This must be why I'm listening to Pokémon OSTs. Childhood nostalgia is the only thing that can save me from this issue now.

> someone to write tests

**Oh! Oh!** That could be *super* awesome. Allow me to saddle you with many regretful burdens when the time comes. :smile:

PR #616, opened 2026-02-25
Make `Any` a subhint of everything

### LINKED ISSUE #530: [DOOR Bug] `beartype.door.is_subhint(Any, str | None) == False` but should be `True` 😮
Courtesy the `typing` guru twins @davidfstr and @erictraut at python/mypy#19227, `beartype.door.is_subhint(Any, str | None)` currently returns `False` but should instead return `True`. Thankfully, this should be trivial for us to correct. Probably *all* calls of the form `is_subhint(Any, T)` for any arbitrary type hint `T` should now unconditionally return `True`.

Why? Allow me to now shamelessly copy-pasta @erictraut, whose [word is law](https://github.com/python/mypy/issues/19227#issuecomment-2949713065):

> The tldr is that `Any` is a stand-in for "some type that is statically unknown that could satisfy assignability rules". That means `Any` is assignable to `str | None` because there exist types in the type system that are so assignable. Likewise, `str | None` is assignable to `Any`. But `set[str | None]` is not assignable to `list[Any]` because there are no types in the type system that would make this assignment legal.

Interestingly, the subhint relation implemented by `is_subhint()` is no longer the cute side curiosity we once thought it was; it's the theoretical formal basis for PEP 747's `TypeForm[...]` type hint factory, whether anybody believes @leycec or not. This issue is thus a dependency of #505, which intends to deeply type-check `TypeForm[...]` type hints at runtime by internally deferring to `is_subhint()`.

Let's do this! Team QA: *Assemble.* 🦸 🦹

Glinte:
Fixes #530

First commit is a surface level fix that simply fixes `is_subhint(Any, int)` but not `is_subhint(list[Any], list[int])`, second commit reverts that, goes deeper and supposedly should fix everything but it touches deeper internals so I have lower confidence.

### DIFF
--- a/beartype/door/_cls/doorsuper.py
+++ b/beartype/door/_cls/doorsuper.py
@@ -713,6 +713,9 @@ class TypeHint(Generic[T_Hint], metaclass=_TypeHintMetaclass):
 
         # Return true only if either...
         return (
+            # This hint is the "typing.Any" catch-all (then this hint is
+            # necessarily a subhint of any hint) *OR*...
+            self._hint is Any or
             # That hint is the "typing.Any" catch-all (then this hint is
             # necessarily a subhint of that hint) *OR*...
             other._hint is Any or
--- a/beartype/door/_cls/pep/pep484/doorpep484typevar.py
+++ b/beartype/door/_cls/pep/pep484/doorpep484typevar.py
@@ -18,7 +18,6 @@ from beartype.door._cls.pep.doorpep484604 import UnionTypeHint
 # from beartype.roar import BeartypeDoorPepUnsupportedException
 from beartype.typing import (
     TYPE_CHECKING,
-    Any,
     TypeVar,
 )
 from beartype._util.cache.utilcachecall import property_cached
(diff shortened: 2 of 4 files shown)


leycec:
**So good!** @Glinte cares about `is_subhint()`. This is wondrous. You might also be interested in assaulting @beartype's [growing laundry list of *other* open `is_subhint()` issues](https://github.com/beartype/beartype/issues/?q=is%3Aissue%20state%3Aopen%20is_subhint), especially the most recent three by @asitstands. I probably won't have time to brutally fix those issues before the @beartype `0.23.0rc0` micdrop in a week or two. If you'd like to throw either Claude or your own Claude-like brilliant mind at any of those, that would be equally wondrous!

**...ugh.** Now I need to man up for some PR code review, huh? But coding is so much funner... :face_exhaling:

leycec:
**Excellente!** Indeed, this block in the initial commit was suspicious:

```python
    # By design, "typing.Any" is assignable to every type hint at this
    # public API boundary. Note that this check intentionally lives here
    # rather than internally in TypeHint.is_subhint(), because unconstrained
    # TypeVars are internally represented as TypeHint(Any) branches --
    # making Any a universal subhint internally would incorrectly cause
    # unconstrained TypeVars to also be subhints of everything.
    if subhint is Any:
        return True
```

...but the followup commit then wisely removed that. Miraculously, everything passes. This is tremendous work. Let us crush #530 with finality and power. Thanks so much for being our perennial Bugfinder General. You crush the years-old issues I no longer have the intestinal strength to understand. Let's get this QA party started. :partying_face:

PR #647, opened 2026-05-18
Does this make Zensical? (docs migration proof-of-concept)

posita:
_**UPDATE to the UPDATE:** Now includes ports of existing, hand-crafted, artisanal API prose docs. This is now a complete port of all existing Sphinx/RTD docs from `doc/src`._

### :point_down: :point_down: :point_down: This is the easiest way to see what this whole mess does. :point_down: :point_down: :point_down: 

To view the latest version locally:

```sh
git clone https://github.com/beartype/beartype.git
cd beartype
git checkout posita/1/bearly-documented-fur-real
uv venv --clear --prompt bearly-documented-fur-real --relocatable  # --or--  python3 -m venv .venv
uv pip install  # --or--  .venv/bin/pip install --editable .[doc-ghp]
.venv/bin/zensical serve  # now go to http://localhost:8000/
```

### :point_up_2: :point_up_2: :point_up_2: Seriously, just do this. You won't regret it. Or you might. Not sure. Either way, you'll have your answer! :point_up_2: :point_up_2: :point_up_2:

_**UPDATE:** Now depends on ~~#651~~ based on https://github.com/beartype/beartype/pull/647#issuecomment-4524689088._

Surfacing this for ~~early~~ feedback. This is _now_ a ~~partial~~ complete_ migration from existing reStructuredText documentation in `doc/src` to ~~MkDocs~~ _Zensical_ Markdown. (No auto-generated API docs yet, but that's a different effort.) This is only content at this point. Nothing structural has changed. (All existing docs and deployment machinery are still in place.)

Mentioned recently in this https://github.com/beartype/beartype/issues/642#issuecomment-4457253610 _et seq._

Possible avenues from here include:

1. ~~SOLD! I'll take it! Merge this, hurry up, and start another one to finish the job! :horse_racing::checkered_flag:~~ _This ship has likely sailed at this point, since I couldn't stop messing with stuff._
2. Yay! ~~Keep going!~~ _You did all the things_ **way** _fast_ <sup>which is terrifying :grimacing:</sup>_!_ We'll merge the whole nine yards later <sup>_maybe, or never</sup> ~~all at once~~! YOLO! :speedboat::parachute:
3. Meh. This is a good start, but you ***really*** should have leaned on all those years of software engineering "experience" and submitted this as multiple chained PRs. Shame. SHAME! 🫵😠
4. Thanks. I'm good. We're good. No thanks. But thanks. :point_up::triumph:
[...]

PR #651, opened 2026-05-23
Don't rock the jukebox: Add Zensical dependencies before we need 'em

posita:
Additional prep for #647 to introduce of Zensical into pyproject.toml as a lightweight strategy sa[gn]ely requested by @leycec at https://github.com/beartype/beartype/pull/647#issuecomment-4524689088.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -209,6 +209,9 @@ dependencies = []
 # hosting service.
 doc-ghp = [
     # ....................{ THEME                          }....................
+    # FIXME: Probably remove this later after committing to Zensical, since it's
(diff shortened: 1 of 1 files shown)


leycec:
Thank you so much for humouring me. This was annoying. I know. Sadly, I know myself. Soon, you will too. I'll procrastinate for what feels to both of us like multiple time-dilated lifetimes on Salvia as an immobile potted ficus named "Betty" before merging #647. It's better this way. :joy: 


<sup>*that moment when you realize being an immobile potted ficus named "Betty" ain't so bad, really*</sup>

PR #667, opened 2026-06-14
Set py-3.15 test targets as experimental for now?

posita:
_UPDATE: The root cause/motivation for this change has been mitigated (see 1ece3478e2c6c3ec96c0b1a9b853ead4bbdf0859), so an effort to paper over new things may be misguided at this point. Feel free to abandon with ... erm ... abandon._

Uses an "experimental" key in matrix jobs with the "continue-on-error" setting to identify test jobs that may fail, but which shouldn't short-circuit other jobs.

Motivation: At some point recently (probably as a result of something I did), test jobs started to fail for Python 3.15 specifically (e.g., [this one](https://github.com/beartype/beartype/actions/runs/27491595337)). I haven't dug in, but superficially, this looks like a Sphinx version peg incompatibility somewhere?

The side effect of a Python 3.15 test job failing is that the failure will short-circuit other test jobs, so one can't easily tell whether it's a universal failure, or limited to the alpha version. This PR provides a mechanism to mark specific Python implementations as "experimental" so their failures don't hide test execution for stable versions.

This does _**NOT**_ address the root cause for the [Python 3.15 failures, which persist](https://github.com/beartype/beartype/actions/runs/27507513252). That ~is~ _was_ a [separate effort](https://github.com/beartype/beartype/commit/1ece3478e2c6c3ec96c0b1a9b853ead4bbdf0859).

### DIFF
--- a/.github/workflows/python_test.yml
+++ b/.github/workflows/python_test.yml
@@ -149,13 +149,17 @@ jobs:
         #   embedded in a single quoted version specifier. For example,
         #   selecting the Python 3.11 pre-release reduces to:
         #     python-version: "3.11.0-alpha - 3.11.0"
+        # * The key/value pair "experimental: true" is used below as
+        #   part of the "continue-on-error" setting so that a failure
+        #   in a job marked as experimental will not short-circuit
+        #   adjacent matrix jobs.
         include:
           - { python-version: "3.10",  tox-env: "py310-coverage" }
           - { python-version: "3.11",  tox-env: "py311-coverage" }
           - { python-version: "3.12",  tox-env: "py312-coverage" }
           - { python-version: "3.13",  tox-env: "py313-coverage" }
           - { python-version: "3.14",  tox-env: "py314-coverage" }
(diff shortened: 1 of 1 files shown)


leycec:
**Superbness!** This is quite a useful CI improvement. The implicit string matching approach is kinda cool, too. Kinda. Also kinda sus. Should one *really* be matching strings in CI runs? Probably. Kinda sus is the best sus. Still, Python keeps telling me that "explicit is better than implicit." Perhaps you were right not to go there. I keep pointing Python at @beartype, though, which keeps saying:

> "B-b-but I'm literally implicit. I'm so implicit I radically transform Python into Bearython with just a single `beartype_this_package()` call. Bearython is Bearable Python. It matters, Python."

Python keeps not caring.

Indeed, @beartype does tend to spuriously fail under Python prereleases. Which is *soooooo* annoying whenever it happens. Thanks to your excellent work, now I can just ignore those failures until people start yelling at me. Even then, of course, I'll still do nothing. Let the people eat the CI cake of anger!


<sup>*@leycec every time he tries eating cake in the family dojo*</sup>

Issue #612, opened 2026-01-17
[Bug] `is_subhint(Child[T, U, V], Parent[T, V])` does not work as expected


asitstands:
It looks like `is_subhint(Child[T, U, V], Parent[T, V])` does not work as expected. Thanks !

```python
from beartype.door import is_subhint

class Parent[T, U]:
    pass

# Parameter positions of Child do not match with Parent.
class Child[T, U, V](Parent[T, V]):
    pass

# Assertion fails
assert is_subhint(Child[str, int, float], Parent[str, float])

# If parameter positions match, it works as expected.
class Child2[T, U, V](Parent[T, U]):
    pass

assert is_subhint(Child2[str, int, float], Parent[str, int])
```

leycec:
**ohnoes!!!!** More horrifying `is_subhint()` use cases involving non-trivial subscripted generic type hierarchies. This is why our cats scream on Saturday evening. 🙀 🙀 🙀

Glinte:
This patch works but I haven't cleaned up the AI code. At least there are some new test cases for more edge cases. Parent commit is `cc57c8311974bf9ce8b8567c7ed05aec856a7827`

```patch
Index: beartype_test/a00_unit/a30_api/door/_fixture/_doorfix_is_subhint.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/beartype_test/a00_unit/a30_api/door/_fixture/_doorfix_is_subhint.py b/beartype_test/a00_unit/a30_api/door/_fixture/_doorfix_is_subhint.py
--- a/beartype_test/a00_unit/a30_api/door/_fixture/_doorfix_is_subhint.py	(revision cc57c8311974bf9ce8b8567c7ed05aec856a7827)
+++ b/beartype_test/a00_unit/a30_api/door/_fixture/_doorfix_is_subhint.py	(date 1772003426787)
@@ -51,7 +51,10 @@
     import typing
     from abc import ABCMeta
     from beartype._util.cls.utilclstest import is_type_subclass_proper
-    from beartype._util.py.utilpyversion import IS_PYTHON_AT_MOST_3_13
+    from beartype._util.py.utilpyversion import (
+        IS_PYTHON_AT_LEAST_3_12,
+        IS_PYTHON_AT_MOST_3_13,
+    )
     from beartype_test.a00_unit.data.pep.generic.data_pep484generic import (
         Pep484GenericT,
[...]

PR #669, opened 2026-06-25
Fix is_subhint() for generics with non-sequential TypeVar ordering in parent subscription

binggao1230:
## Summary

Fixes `is_subhint()` incorrectly returning `False` when a child class uses its type parameters in a non-sequential order when subscripting a parent class.

**Repro** (reported as #612):
```python
from beartype.door import is_subhint
from typing import TypeVar, Generic

T = TypeVar('T')
U = TypeVar('U')
V = TypeVar('V')

class Parent(Generic[T, U]):
    pass

class Child(Parent[T, V], Generic[T, U, V]):
    pass

# Before fix: returns False (wrong)
# After fix:  returns True  (correct)
assert is_subhint(Child[str, int, float], Parent[str, float])
```

## Root Cause

In `get_hint_pep484585_generic_args_full()`, the depth-first search (DFS) "bubbles up" type parameters from child generics into parent generics using a sequential stack-pop mechanism. When `Child[str, int, float]` is traversed (with `Child.__parameters__ == (T, U, V)` so `T→str, U→int, V→float`), the DFS encounters TypeVars `T` and `V` (in that order) from `Parent[T, V]` being bubbled back to `Child`. Sequential stack-popping assigned:
- `T` → `str` (1st pop) ✓
- `V` → `int` (2nd pop) ✗ — should be `float` (position 2 in `__parameters__`)

## Fix

Before the DFS loop, build an authoritative `TypeVar → concrete arg` mapping from the root generic's `__parameters__` and subscription args (`hint_root_typearg_to_hint`). When bubbling TypeVars directly back into the root generic's args-stack, consult this mapping instead of the sequential stack-pop. This correctly resolves `V → float` by position, regardless of DFS encounter order.

The fix is minimal and surgical: it only applies when (a) we're bubbling into the root generic and (b) the TypeVar appears in the root's own `__parameters__`.

## Test Plan

- [ ] New `Pep484GenericSTU_ST` test generic added to `data_pep484generic.py`
- [ ] Five new `is_subhint` test cases covering the bug fix and negative cases
- [ ] All 416 existing passing tests continue to pass (`python3 -m pytest beartype_test/ -q`)

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/beartype/_util/hint/pep/proposal/pep484585/generic/pep484585genget.py
+++ b/beartype/_util/hint/pep/proposal/pep484585/generic/pep484585genget.py
@@ -227,6 +227,53 @@ def get_hint_pep484585_generic_args_full(
         is_hint_pep484585_generic_user)
     from beartype._util.hint.pep.utilpepget import get_hint_pep_args
 
+    # ....................{ LOCALS ~ typevar ~ root        }....................
+    # Direct mapping from each type parameter (i.e., "typing.TypeVar" object)
+    # of the unsubscripted root generic to the corresponding concrete child hint
+    # directly subscripting the passed root generic, built from:
+    # * The "__parameters__" attribute of the unsubscripted root generic (i.e.,
+    #   the ordered tuple of all type parameters that generic is parametrized by).
+    # * The args directly subscripting the passed (possibly subscripted) root
+    #   generic (i.e., the concrete hints filling those type parameters).
+    #
+    # This mapping is the authoritative TypeVar -> concrete hint lookup for the
+    # root generic. The DFS "bubbling up" mechanism may assign incorrect values
+    # to "typearg_to_hint" when a child class uses its type parameters in a
+    # non-sequential order when subscripting a parent class: e.g.,
+    #     >>> class Parent(Generic[T, U]): pass
+    #     >>> class Child(Parent[T, V], Generic[T, U, V]): pass
+    #     >>> get_hint_pep484585_generic_args_full(Child[str, int, float], Parent)
+    #     (str, float)   # T=str, V=float, NOT (str, int)
+    #
+    # When the DFS encounters the TypeVar "V" bubbled up through "Parent[T, V]",
+    # sequential popping from Child's args_stack incorrectly maps V to "int"
+    # (the second arg) rather than "float" (the third arg). This mapping
+    # corrects that by providing the authoritative V -> float mapping.
+    hint_root_args = get_hint_pep_args(hint)
+    hint_root_params = getattr(
+        get_hint_pep484585_generic_type(  # pyright: ignore
+            hint=hint,
+            exception_cls=exception_cls,
+            exception_prefix=exception_prefix,
+        ),
+        '__parameters__', ()
+    )
+
+    # Build the direct TypeVar -> concrete hint mapping for the root generic,
+    # but ONLY for type parameters paired with concrete (non-TypeVar) args.
+    # Mapping a TypeVar to another TypeVar would pollute this mapping and inhibit
+    # subsequent resolution of that TypeVar to a concrete hint.
+    hint_root_typearg_to_hint: Pep484612646TypeArgUnpackedToHint = (
+        {
+            hint_root_param: hint_root_arg  # type: ignore[misc]
+            for hint_root_param, hint_root_arg in zip(
+                hint_root_params, hint_root_args)
+            if not is_hint_pep484612646_typearg_unpacked(hint_root_arg)
+        }
+        if hint_root_args and len(hint_root_params) == len(hint_root_args) else
+        {}
+    )
+
     # ....................{ PREAMBLE                       }....................
     # If the caller explicitly passed a pseudo-superclass target...
     #
@@ -539,9 +586,46 @@ def get_hint_pep484585_generic_args_full(
                             elif hint_base_super_args_stack:
                                 # print(f'Bubbling hint {hint_base_super_args_stack[-1]} into...')
                                 # print(f'base {hint_base} typevar {hint_base_arg}!')
-                                hint_base_arg_full_new = (  # pyright: ignore
-                                hint_base_args_full[hint_base_arg_full_index]) = (
-                                    hint_base_super_args_stack.pop())  # type: ignore[assignment]
+                                #
+                                # If we are bubbling this type parameter directly
+                                # into the root generic AND the root generic's
+                                # authoritative TypeVar mapping already covers
+                                # this type parameter, use that mapping rather
+                                # than sequentially popping from the root's
+                                # args-stack. Sequential popping is incorrect
+                                # when a child class uses its type parameters in
+                                # a non-sequential order when subscripting a
+                                # parent class: e.g.,
+                                #     class Parent(Generic[T, U]): ...
+                                #     class Child(Parent[T, V], Generic[T, U, V]):
+                                #         ...
+                                # In Child[str, int, float], the DFS encounters
+                                # TypeVars T and V (in that order) when bubbling
+                                # from Parent[T,V] back to Child. Sequential
+                                # popping maps V -> "int" (the 2nd arg), but the
+                                # correct mapping via __parameters__ is V ->
+                                # "float" (the 3rd arg).
+                                if (
+                                    hint_base_super_data is hint_root_data and  # pyright: ignore
+                                    hint_base_arg_full in hint_root_typearg_to_hint
+                                ):
+                                    # Use the authoritative root mapping, which
+                                    # correctly resolves TypeVars by their
+                                    # position in the root generic's
+                                    # __parameters__ rather than by the encounter
+                                    # order of the DFS traversal.
+                                    # Do NOT pop from the args-stack: the stack
+                                    # may still be needed for other TypeVars
+                                    # whose concrete values come from inner-level
+                                    # subscriptions not present in the root's
+                                    # own __parameters__.
+                                    hint_base_arg_full_new = (  # pyright: ignore
+                                    hint_base_args_full[hint_base_arg_full_index]) = (
+                                        hint_root_typearg_to_hint[hint_base_arg_full])  # type: ignore[index]
+                                else:
+                                    hint_base_arg_full_new = (  # pyright: ignore
+                                    hint_base_args_full[hint_base_arg_full_index]) = (
+                                        hint_base_super_args_stack.pop())  # type: ignore[assignment]
 
                                 # If the currently unassigned child hint
                                 # directly subscripting this parent
@@ -702,18 +786,36 @@ def get_hint_pep484585_generic_args_full(
         for hint_arg_full_index, hint_arg_full in enumerate(hint_args_full):
             # If this hint is a type parameter...
             if is_hint_pep484612646_typearg_unpacked(hint_arg_full):
-                # Either:
-                # * If a child hint directly subscripting a sibling
-                #   pseudo-superclass of this target pseudo-superclass has
-                #   already been "bubbled up" into this type parameter, preserve
-                #   that bubbling by re-bubbling up the same child hint back
-                #   into this # type parameter. <-- lol
-                # * If *NO* child hint directly subscripting a sibling
-                #   pseudo-superclass of this target pseudo-superclass has
-                #   already been "bubbled up" into this type parameter, preserve
-                #   this type parameter as is.
-                hint_args_full[hint_arg_full_index] = typearg_to_hint.get(  # type: ignore[call-overload]
-                    hint_arg_full, hint_arg_full)  # type: ignore[arg-type]
+                # Resolve this type parameter to a concrete hint by consulting,
+                # in priority order:
+                # 1. The authoritative "hint_root_typearg_to_hint" mapping,
+                #    which directly maps each type parameter of the root generic
+                #    to its concrete subscription arg. This mapping is consulted
+                #    FIRST because it is always correct: e.g., in
+                #    "Child[str, int, float]" with "Child.__parameters__ ==
+                #    (T, U, V)", this mapping contains {T: str, U: int, V:
+                #    float}. The DFS "bubbling up" mechanism may corrupt
+                #    "typearg_to_hint" with incorrect values when a child class
+                #    uses its type parameters in a non-sequential order when
+                #    subscripting a parent class (e.g., "class Child(Parent[T,
+                #    V], Generic[T, U, V])": sequential popping maps V to "int"
+                #    rather than "float"), so we prefer this mapping.
+                # 2. The DFS-accumulated "typearg_to_hint" mapping, which
+                #    captures type parameters bubbled up through sibling
+                #    pseudo-superclasses. It covers type parameters of
+                #    intermediate generics not present in the root generic's
+                #    own "__parameters__".
+                # 3. The type parameter itself (i.e., no resolution), when
+                #    neither mapping has an entry for this type parameter.
+                hint_args_full[hint_arg_full_index] = (
+                    hint_root_typearg_to_hint.get(  # type: ignore[call-overload]
+                        hint_arg_full,
+                        typearg_to_hint.get(  # type: ignore[call-overload]
+                            hint_arg_full,
+                            hint_arg_full,  # type: ignore[arg-type]
+                        ),
+                    )
+                )
             # Else, this hint is *NOT* a type parameter.
     # Else, the caller did *NOT* pass a target pseudo-superclass.
     else:
--- a/beartype_test/a00_unit/a30_api/door/_fixture/_doorfix_is_subhint.py
+++ b/beartype_test/a00_unit/a30_api/door/_fixture/_doorfix_is_subhint.py
@@ -57,6 +57,7 @@ def door_cases_is_subhint() -> 'Iterable[Tuple[object, object, bool]]':
         Pep484GenericSubT,
         Pep484GenericST,
         Pep484GenericSInt,
+        Pep484GenericSTU_ST,
     )
     from beartype_test.a00_unit.data.pep.generic.data_pep585generic import (
         Pep585SequenceT,
@@ -273,6 +274,19 @@ def door_cases_is_subhint() -> 'Iterable[Tuple[object, object, bool]]':
         (Pep484GenericSInt[str], Pep484GenericST[T_sequence, S], True),
         (Pep484GenericSInt[Sequence], Pep484GenericST[list, object], False),
         (Pep484GenericSInt[T_sequence], Pep484GenericST, True),
+
+        # PEP 484-compliant generic subclasses parametrized by three type
+        # variables but subscripting a two-parameter parent in non-sequential
+        # order (i.e., first and third type parameters, skipping the second).
+        # This exercises the fix for GitHub issue #612, where is_subhint()
+        # incorrectly mapped type parameters by sequential encounter order
+        # during the depth-first search traversal rather than by their
+        # authoritative position in the root generic's __parameters__.
+        (Pep484GenericSTU_ST, Pep484GenericST, True),
+        (Pep484GenericSTU_ST[str, int, float], Pep484GenericST, True),
+        (Pep484GenericSTU_ST[str, int, float], Pep484GenericST[str, float], True),
+        (Pep484GenericSTU_ST[str, int, float], Pep484GenericST[str, int], False),
+        (Pep484GenericSTU_ST[str, int, float], Pep484GenericST[str, str], False),
 
         # ..................{ PEP 484 ~ optional             }..................
         # "typing.Optional"-centric tests.
--- a/beartype_test/a00_unit/data/pep/generic/data_pep484generic.py
+++ b/beartype_test/a00_unit/data/pep/generic/data_pep484generic.py
@@ -15,6 +15,8 @@ from beartype_test.a00_unit.data.data_type import Class
 from beartype_test.a00_unit.data.pep.pep484.data_pep484 import (
     S,
     T,
+    U,
+    V,
 )
 from collections.abc import (
     Sized as SizedABC,
@@ -119,6 +121,23 @@ class Pep484GenericIntInt(Pep484GenericST[int, int]):
     :pep:`484`-compliant concrete generic subclass inheriting a
     :pep:`484`-compliant generic superclass subscripted twice by the
     builtin :class:`int` type.
+    '''
+
+    pass
+
+
+class Pep484GenericSTU_ST(Pep484GenericST[S, U], Generic[S, T, U]):
+    '''
+    :pep:`484`-compliant generic subclass parametrized by three type variables
+    but inheriting a :pep:`484`-compliant generic superclass subscripted by
+    only two of those type variables in **non-sequential** order (i.e., the
+    first and third type variables ``S`` and ``U``, skipping the second ``T``).
+
+    This class exercises the edge case where a child class uses its type
+    parameters in a non-sequential order when subscripting a parent class,
+    requiring :func:`beartype.door.is_subhint` to correctly resolve type
+    parameters by their position in ``__parameters__`` rather than by the
+    encounter order of the depth-first search traversal.
     '''
 
     pass


leycec:
**Brilliance.** I was desperately hoping a brave shining code knight would do this for the good of me and QA. And, lo! My saviour appeared. Apparently, Brave Sir @gaoflow is my saviour. You and your beautiful tabby cats have spared me the spiritual anguish of attempting to understand the unintelligible `get_hint_pep484585_generic_args_full()` algorithm at my advanced mental age. :joy: 

These changes are... intense. Of course, tests pass. I have no doubt these changes are correct and useful. Still, I ask you to *bear* <sup>hyuk hyuk</sup> with me for a day or two as I sift through everything line by line. Even if it helps nothing, I'd still like to understand exactly what I'm agreeing to here.

Thank you so much for this awesome resolution, Brave Sir @gaoflow. I'm eternally grateful to both you and the genius-tier LLM you prompted. I hope you're having an awesome summer, too. You rock! :guitar: :notes:

beartype/_util/hint/pep/proposal/pep484585/generic/pep484585genget.py
```diff
@@ -227,6 +227,53 @@ def get_hint_pep484585_generic_args_full(
...
+    #     >>> class Child(Parent[T, V], Generic[T, U, V]): pass
+    #     >>> get_hint_pep484585_generic_args_full(Child[str, int, float], Parent)
+    #     (str, float)   # T=str, V=float, NOT (str, int)
+    #
+    # When the DFS encounters the TypeVar "V" bubbled up through "Parent[T, V]",
+    # sequential popping from Child's args_stack incorrectly maps V to "int"
+    # (the second arg) rather than "float" (the third arg). This mapping
+    # corrects that by providing the authoritative V -> float mapping.
+    hint_root_args = get_hint_pep_args(hint)
+    hint_root_params = getattr(
```
leycec, line 253: **Tragedy strikes!** Manually introspecting type parameters by calling `getattr(..., '__parameters__', ()) isn't a great idea, really. Doing so breaks down in common edge cases like PEP 646-compliant unpacked type variable tuples, which CPython (uselessly and thus basically erroneously) stores in *packed* rather than unpacked form in the `__parameters__` dunder attribute. Instead, consider calling our high-level `get_hint_pep_typeargs_unpacked()` getter: e.g.,

```python
from beartype._util.hint.pep.utilpepget import get_hint_pep_typeargs_unpacked
...

    #FIXME: Improve the get_hint_pep_typeargs_unpacked() getter to
    #internally perform this call to the get_hint_pep484585_generic_type()
    #getter. Doing so is necessary to circumvent flaws in the low-level
    #"__parameters__" dunder attribute, as also documented by similar
    #"#FIXME:" comments preceding get_hint_pep_typeargs_packed(). In short,
    #this should just be automated away already. Too much generics madness!
    hint_root_unsubscripted = get_hint_pep484585_generic_type(  # pyright: ignore
        hint=hint,
        exception_cls=exception_cls,
        exception_prefix=exception_prefix,
    )
    hint_root_params = get_hint_pep_typeargs_unpacked(hint_root_unsubscripted)
```

Pretty sure that should make everything worky. Pretty sure. But not certain. Let's pretend I actually know what I'm talking about. 😅 

**P.S.:** Oh, and don't worry about the new `#FIXME:` comment I injected above. That's just a note to my future self to resolve a really dumb long-standing issue. More sweating emoji guy. :cold_sweat: :sweat: :sweat_smile:

leycec:
More incisive comments that possibly mean something to follow on Monday! Looks great so far. Superb volunteerism, everybody. Have a great weekend and may the QA be with us all. :smile:

beartype/_util/hint/pep/proposal/pep484585/generic/pep484585genget.py
```diff
@@ -227,6 +227,53 @@ def get_hint_pep484585_generic_args_full(
...
+    # Mapping a TypeVar to another TypeVar would pollute this mapping and inhibit
+    # subsequent resolution of that TypeVar to a concrete hint.
+    hint_root_typearg_to_hint: Pep484612646TypeArgUnpackedToHint = (
+        {
+            hint_root_param: hint_root_arg  # type: ignore[misc]
+            for hint_root_param, hint_root_arg in zip(
+                hint_root_params, hint_root_args)
+            if not is_hint_pep484612646_typearg_unpacked(hint_root_arg)
+        }
+        if hint_root_args and len(hint_root_params) == len(hint_root_args) else
```
leycec, line 273: Interestingly, this conditional efficiently reduces to just:

```python
if len(hint_root_params) == len(hint_root_args) else
```

Why? Because `len(hint_root_params) > 0` is guaranteed to *always* be the case. Why? Because all generics are necessarily parametrized by at least one type variable.

Likewise, rather than noop out to a new empty dictionary `{}`, we can efficiently noop out to the @beartype-specific empty frozen dictionary singleton `FROZENDICT_EMPTY`. Congealing these two optimizations together then yields:

```python
from beartype._data.kind.datakindmap import FROZENDICT_EMPTY
...

        if len(hint_root_params) == len(hint_root_args) else
        FROZENDICT_EMPTY
```

That said, I'm unconvinced that only conditionally defining a non-empty `hint_root_typearg_to_hint` dictionary when `len(hint_root_params) == len(hint_root_args)` is the correct decision. That doesn't seem right at all. Shouldn't we just *always* unconditionally define this dictionary? Technically, it's true that we only need to define this dictionary if `len(hint_root_params) >= 2`. (Why? Because a generic parametrized by only one type variable obviously has no ordering issues.) Pragmatically, though, trying to micro-optimize this non-empty dictionary away into an empty dictionary under various edge cases just increases the likelihood of erroneous results. Speed is important. Correctness is more important. This algorithm is brutal enough, people! I can hardly bare to gaze upon its hideous visage even in its current state. 😄 

Lastly, the local variable name `hint_root_typearg_to_hint` is tragically ambiguous. Type arguments are themselves valid hints, whereas this dictionary explicitly maps type arguments to non-type arguments. Although verbose, a local variable name of something like `hint_root_typearg_to_nontypearg` would be more accurate.

**Note to self:** for this same reason, we should also rename the local `typearg_to_hint` dictionary to `typearg_to_nontypearg`. `</shrug_shrug>`

In other words, I'm pretty sure this entire variable assignment should just be reduced to this dictionary comprehension:

```python
hint_root_typearg_to_nontypearg: Pep484612646TypeArgUnpackedToHint = {
    hint_root_param: hint_root_arg  # type: ignore[misc]
    for hint_root_param, hint_root_arg in zip(
        hint_root_params, hint_root_args)
    if not is_hint_pep484612646_typearg_unpacked(hint_root_arg)
}
```

**...gods.** Pretty intense stuff, huh? My GitHub avatar is sweating. 😅

beartype/_util/hint/pep/proposal/pep484585/generic/pep484585genget.py
```diff
@@ -539,9 +586,46 @@ def get_hint_pep484585_generic_args_full(
...
+                                    # correctly resolves TypeVars by their
+                                    # position in the root generic's
+                                    # __parameters__ rather than by the encounter
+                                    # order of the DFS traversal.
+                                    # Do NOT pop from the args-stack: the stack
+                                    # may still be needed for other TypeVars
+                                    # whose concrete values come from inner-level
+                                    # subscriptions not present in the root's
+                                    # own __parameters__.
+                                    hint_base_arg_full_new = (  # pyright: ignore
```
leycec, line 622: Oh, frabjous day! Let's avoid assigning `hint_base_arg_full_new` here. That local variable is only of relevance to the following `else:` branch. It's not much of a micro-optimization, but I'll take it. 🥳 

**UPDATE:** Lolbro. Nevermind! Per prior commentary, we (probably) do need to explicitly find and remove `hint_root_typearg_to_hint[hint_base_arg_full]` from `hint_base_super_args_stack`. Ergo, we (probably) do still need this `hint_base_arg_full_new` local variable to make that happen. `</shrug_shrug_shrug>`

beartype/_util/hint/pep/proposal/pep484585/generic/pep484585genget.py
```diff
@@ -539,9 +586,46 @@ def get_hint_pep484585_generic_args_full(
...
+                                if (
+                                    hint_base_super_data is hint_root_data and  # pyright: ignore
+                                    hint_base_arg_full in hint_root_typearg_to_hint
+                                ):
+                                    # Use the authoritative root mapping, which
+                                    # correctly resolves TypeVars by their
+                                    # position in the root generic's
+                                    # __parameters__ rather than by the encounter
+                                    # order of the DFS traversal.
+                                    # Do NOT pop from the args-stack: the stack
```
leycec, line 617: **...hmm.** I get why we're intentionally not performing a `hint_base_super_args_stack.pop()` in this edge case. That said, we (probably) *should* still be explicitly finding and removing `hint_root_typearg_to_hint[hint_base_arg_full]` (i.e., the concrete child hint subscripting the root generic that maps to the current type parameter transitively parametrizing the current child pseudo-superclass) from `hint_base_super_args_stack`. It's (probably) not okay to just leave concrete child hints subscripting the root generic that have already been mapped to type parameters as vestigial relics in `hint_base_super_args_stack`. That stack means something. Disrupting its internal semantics obstructs that meaning. That way lies meaningless madness, which is what my life already suffers from. 😂 -> 😭 

I'll also be the first to unhappily admit I have no idea exactly what minimal reproducible example (MRE) would reliably reproduce this badness. Still, badness is (probably) happening here.

Obviously, tests currently pass. But that (probably) just means that tests are currently deficient. Which... totally makes sense. I barely understood this algorithm even when I wrote it. Since so much "probably" is happening above, it's clear I now understand this algorithm even less than I once did. 🥹 -> 🥲

beartype/_util/hint/pep/proposal/pep484585/generic/pep484585genget.py
```diff
@@ -702,18 +786,36 @@ def get_hint_pep484585_generic_args_full(
...
+                #    to its concrete subscription arg. This mapping is consulted
+                #    FIRST because it is always correct: e.g., in
+                #    "Child[str, int, float]" with "Child.__parameters__ ==
+                #    (T, U, V)", this mapping contains {T: str, U: int, V:
+                #    float}. The DFS "bubbling up" mechanism may corrupt
+                #    "typearg_to_hint" with incorrect values when a child class
+                #    uses its type parameters in a non-sequential order when
+                #    subscripting a parent class (e.g., "class Child(Parent[T,
+                #    V], Generic[T, U, V])": sequential popping maps V to "int"
+                #    rather than "float"), so we prefer this mapping.
```
leycec, line 802: **[Full context]** The distressing new comment here reads:

```python
                #    float}. The DFS "bubbling up" mechanism may corrupt
                #    "typearg_to_hint" with incorrect values when a child class
                #    uses its type parameters in a non-sequential order when
                #    subscripting a parent class (e.g., "class Child(Parent[T,
                #    V], Generic[T, U, V])": sequential popping maps V to "int"
                #    rather than "float"), so we prefer this mapping.
```

**...uhh.** What? Really? The local `typearg_to_hint` dictionary still gets corrupted even after applying this changeset? If true, that's both bad and surprising. Like, super-hyper-bad and super-hyper-surprising. That's a significant issue. If `typearg_to_hint` *ever* ends up in a corrupted state, this algorithm will *absolutely* return erroneous results in a subset of valid use cases. In other words, `typearg_to_hint` should *never* end up in a corrupted state. If that's actually happening, we need to prevent that from happening with even more tedious improvements that break our fragile brainpans rather than crudely and ineffectively hacking around that here. 

**tl;dr:** `typearg_to_hint` needs to be the definitive non-corrupt mapping. This well-intended attempt to circumvent that (possible) corruption here only superficially papers over that issue while failing to actually resolve the underlying badness. The road to QA Hell is paved with well-intended attempts. 😹

leycec:
**Commentary done.** I spent my entire allotment of scarce open-source volunteerism tonight reviewing this – but it was time well-spent. Most of these changes are excellent. Some of these changes are less excellent. In fact, some of these changes smell. I acknowledge that this is mostly my fault. This algorithm is extremely brutal. The comments littering this algorithm are extremely problematic and should be fixed by me, which only increases the brutality. Everywhere I wrote "bubble up" in a comment, for example, should definitely instead have read "bubble down" instead.

I kinda have an intuition of where I should go with this:

1. Probably, **I should just merge this as is.** Tests still pass, which is ultimately all that *really* matters. This is esoteric stuff that most @beartype users are indifferent to. If something breaks as a consequence of merging this, that breakage (probably) only affects a small portion of the userbase. Not ideal – but hardly the end of @beartype.
2. Definitely, **I should then personally attend to all of the comments I made above.** I'm genuinely curious about all of this stuff, especially that bit about `typearg_to_hint` being corrupted. I'd like to see if that's genuinely the case. That *shouldn't* be the case anymore, but... who knows? This algorithm's brutality makes it almost impossible to understand by mere inspection alone. At some point, a mountain of `print()` statements and a week fighting with the debugger is all you can hope for. :weary: 

So... that's kinda the plan. What a plan, huh? Apparently, the plan is now: "Merge it, make a few incoherent changes, and beg the Universe to forgive us all our QA sins." That's not a plan. That's a galactic plea for help. :joy:

Glinte:
I just realize that I have an AI generated patch at https://github.com/beartype/beartype/issues/612#issuecomment-3957299650, even if we ignore the AI implementation, tests with the modern 3.12+ syntax are definitely needed.

leycec:
**OMG!!!** I totally missed that. I'm so sorry, @Glinte. February 25th, huh? Well, isn't that a rusty kettle full of QA fish.

Issue #612 has become a stunning public indictment of my frolicsome nature (and the unending @beartype `0.23.0` dev hell produced by that nature). We now have two competing AI-generated solutions, neither of which can agree on anything and one of which I already exploded three hours at last night. This is a grim glimpse of life in the GitHub trenches.

Oh, well. What can you do except explode three more? :joy:

leycec:
## New Plan Is Good Plan

My new plan is less sucky than my old plan. That's good, because basically there was no old plan. Behold! New Plan™:

1. Manually merge all **unit test changes** from both this PR and [@Glinte's manual changeset](https://github.com/beartype/beartype/issues/612#issuecomment-3957299650).
1. Watch **tests locally break**.
1. Comment out all **unit test changes** until tests locally pretend to pass again.
1. **Commit! Push!** Pray so hard for better QA days.
1. Watch this PR explode with **merge conflicts**. That's not great. Given that @Glinte has a competing solution that looks completely different, though, I'm now staring down the `git` gun of manually merging everything *anyway* by carefully picking and choosing changes between these two changesets. :face_exhaling: 
1. Uncomment those **unit test changes** locally again.
1. Carefully **pick and choose changes** between these two changesets until tests locally pass.
1. **Commit! Push!** We now have succeeded in repairing the fragile fabric of the space-time continuum. Yay!
1. Close out both this PR *and* issue #612 as **resolved**.

There's lots of smelly stuff in both PRs. They can't agree on anything. @Glinte's, for example, is obsessed with the passed `hint_base_target` parameter. Meanwhile, this PR quietly ignores `hint_base_target` altogether. @Glinte's changeset insists that:

```python
    # Pre-seeding is only required when searching for a target pseudo-
    # superclass (i.e., when contextual substitutions are being resolved for
    # subhint comparison).
```

Is that actually true, though? No idea. Probably not. On the one hand, that's suspicious. On the other hand, I like that @Glinte's changeset only requires a `hint_typeargs_preseeded` set rather than this PR's full-blown `hint_root_typearg_to_hint` dictionary. That seems nice. The words "seems" and "nice" are working overtime here.

I don't yet have a clear sense of which PR is likely to win. It might be neither. It might be the human brain that eventually triumphs over the hostile forces of Python's crude `typing` nature. Stay tuned as @leycec actually does something – even if doing something gets us no closer to @beartype `0.23.0`. :joy: -> :tired_face:

leycec:
Unit tests: **merged.**
Meaningful documentation that actually explains this brutalist algorithm: **written.**

Feast your eyes on ASCII art that scours the eyes clean of hope and beauty, QA brethren:

```python
    # ....................{ DESCRIPTION                    }....................
    # This function implements an extremely non-trivial depth-first search (DFS)
    # over all the abstract n-ary tree of all PEP 484- and 585-compliant generic
    # transitive pseudo-superclasses of the passed generic subclass.
    #
    # Semantically, this DFS performs a recursive traversal over this tree.
    # Syntactically, this DFS performs non-recursive iteration only simulating
    # such recursive traversal. While less intuitive (and debuggable,
    # maintainable, and readable), non-recursive iteration is substantially more
    # performant in general and enables even greater performance gains via
    # short-circuiting in this specific case. Due to the extreme
    # non-trivialities involved, we now define terms for sanity and goodwill:
    #
    # * DFS: A naturally recursive form of search typically optimized into a
    #   non-recursive iterative implementation leveraging a Last-In-First-Out
    #   (LIFO) stack. Pure-Python lists define all requisite methods needed to
    #   natively express a LIFO stack and are thus the natural builtin
    #   containers for implementing a DFS non-recursively in Python. Doing so
    #   requires semantically (and usually syntactically) reversing the items
    #   of this list such that the "top" of the LIFO stack expressed by this
    #   list is the last item of this stack. These methods thus include:
    #   * append(), pushing an item into the top of this stack.
    #   * extend(), pushing zero or more items into the top of this stack.
    #   * pop(), popping an item from the top of this non-empty stack.
    # * n-ary tree: A graph-like recursive data structure comprised of nodes and
    #   directed edges between nodes rooted at a single node such that nodes
    #   referred to as:
    #   * Leaves terminate this recursive data structure by originating *NO*
    #     directed edges to other nodes.
    #   * Non-leaves perpetuate this recursive data structure by originating "n"
    #     directed edges to other nodes deeper in the tree, where "n" is a
    #     variable non-zero integer specific to each node.
    # * Pseudo-superclass: Any user-defined PEP 484- and 585-compliant generic
    #   subclasses at least one pseudo-superclass. Since the superclasses that a
    #   generics subclasses are resolved dynamically at inheritance-time via a
    #   non-trivial process referred to as "type erasure," the superficial
    #   subclasses that a generic syntactically appears to subclass may differ
    #   from the actual subclasses that a generic semantically subclasses. For
    #   example, consider this Python snippet defining a type hierarchy of two
    #   non-generics and four generics:
    #       from typing import Generic
    #       class Y(object): ...
    #       class Z(object): ...
    #       class A[S](Z, Generic[S]): ...
    #       class B[T](Y, Generic[T]): ...
    #       class C[S, T](A[S], B[T]): ...
    #       class D[S, T, U](list[T], C[S, T]): ...
    #
    #   These four generics subclass these pseudo-superclasses:
    #   * Generic "A" subclasses both:
    #     * Superclass "Z", which is a normal class (rather than a generic) and
    #       thus an actual superclass of "A" rather than merely a
    #       pseudo-superclass of "A".
    #     * Pseudo-superclass "Generic[S]", which is a subscripted generic and
    #       thus *NOT* an actual superclass of "A". Rather, a private metaclass
    #       of "A" defined by the standard "typing" module silently erases this
    #       pseudo-superclass at inheritance time and permanently replaces this
    #       pseudo-superclass by the normal class "typing.Generic". Ergo,
    #       "Generic[S]" is merely a pseudo-superclass. Syntactically,
    #       "Generic[S]" appears to be a real superclass. Semantically,
    #       "Generic[S]" only masquerades as a real superclass.
    #   * Generic "B" subclasses both superclass "Y" and pseudo-superclass
    #     "Generic[T]", for the exact same reasons.
    #   * Generic "C" subclasses both pseudo-superclasses "A[S]" and "B[T]".
    #   * Generic "D" subclasses both pseudo-superclasses "list[T]" and
    #     "C[S, T]".
    #
    #   This algorithm is principally interested in pseudo-superclasses rather
    #   than the actual superclasses those generics were erased to.
    # * Transitive pseudo-superclass: Any pseudo-superclass anywhere in the
    #   method-resolution order (MRO) of a user-defined PEP 484- and
    #   585-compliant generic. Reusing the above example, the four generics
    #   defined above subclass these transitive pseudo-superclasses:
    #   * Generic "A" transitively subclasses "Z", "Generic[S]", and "object".
    #   * Generic "B" transitively subclasses "Y", "Generic[T]", and "object".
    #   * Generic "C" transitively subclasses "A[S]", "Z", "Generic[S]", "B[T]",
    #     "Y", "Generic[T]", and "object".
    #   * Generic "D" transitively subclasses "list[T]", "C[S, T]", "A[S]", "Z",
    #     "Generic[S]", "B[T]", "Y", "Generic[T]", and "object".
    # * Transitive pseudo-superclass n-ary tree: The n-ary tree rooted at any
    #   user-defined PEP 484- and 585-compliant generic expressing the
    #   relationship of that generic to its transitive pseudo-superclasses. This
    #   tree corresponds to the standard understanding of the relationship
    #   between a subclass and its superclasses, with each subclass treated as a
    #   "child" layered underneath each superclass treated as a "parent".
    #   Reusing the above example, the n-ary tree rooted at generic "D"
    #   expresses the relationship of that generic to its 9 transitive
    #   pseudo-superclasses. In plaintext ASCII, this tree could be depicted as:
    #                               object object object object
    #                                  |     |       |     |
    #                                  Z Generic[S]  Y Generic[T]
    #                                  \ /           \ /
    #                           object A[S]    -----B[T]
    #                              |      \   /
    #                           list[T] C[S, T]
    #                                 \ /
    #                                  D
    # * DFS traversal tree: The same exact tree as the transitive
    #   pseudo-superclass n-ary tree rooted at any user-defined PEP 484- and
    #   585-compliant generic, except depicted such that that generic is the
    #   top-most root node of that tree and each transitive pseudo-superclass of
    #   that generic is a transitive child node of that root node. Although
    #   confusing, this depiction is the "natural" way to express the traversal
    #   order of a DFS over a type hierarchy. Although we typically conceive of
    #   subclasses as "children" of "parent" superclasses, a DFS that traverses
    #   over a type hierarchy asks that we instead momentarily conceive of
    #   subclasses as "parents" of "children" superclasses with respect to
    #   traversal order. Reusing the above example, the DFS traversal rooted at
    #   generic "D" expresses the order in which a DFS traverses over that
    #   generic and its 9 transitive pseudo-superclasses. In plaintext ASCII,
    #   this traversal could be depicted as:
    #                                  D
    #                                 / \
    #                           list[T] C[S, T]
    #                              |      /   \
    #                           object A[S]    -----B[T]
    #                                  / \           / \
    #                                  Z Generic[S]  Y Generic[T]
    #                                  |     |       |     |
    #                               object object object object
    # * Child pseudo-superclass: A pseudo-superclass subclassed by the generic
    #   currently being visited by this DFS. This nomenclature is intrinsic to
    #   the action of a DFS but somewhat counterintuitive from the perspective
    #   of subclass inheritance, where we typically conceive of a subclass as
    #   instead the "child" of a "parent" superclass. Reusing the above example,
    #   "list[T]" and "C[S, T]" are child pseudo-superclasses of generic "D".
    # * Parent pseudo-superclass: A generic or pseudo-superclass subclassing the
    #   generic or pseudo-superclass currently being visited by this DFS. Again,
    #   this nomenclature is intrinsic to the action of a DFS but somewhat
    #   counterintuitive. Reusing the above example, generic "D" is the parent
    #   pseudo-superclass of child pseudo-superclasses "list[T]" and "C[S, T]".
    # * Unpacked type parameter: An abstract type hint to be replaced by a
    #   concrete type hint through the bubbling performed by this DFS.
    #   Specifically, either:
    #   * A PEP 484-compliant type variable (i.e., "typing.TypeVar" object).
    #   * A PEP 612-compliant parameter specification (i.e., "typing.ParamSpec"
    #     object unpacked by the unary "*" prefix operator). For example, the
    #     parameter specification "*P" where "P = typing.ParamSpec('P')".
    #   * A PEP 646-compliant tuple type variable (i.e., "typing.TupleTypeVar"
    #     object unpacked by the unary "*" prefix operator). For example, the
    #     tuple type variable "*Ts" where "Ts = typing.TupleTypeVar('Ts')".
    # * Concrete type hint: Any type hint subscripting a generic such that that
    #   type hint is *NOT* an unpacked type parameter.
    # * Bubbling: The process by which this DFS recursively replaces each
    #   unpacked type parameter parametrizing the target pseudo-superclass of
    #   the passed generic with the corresponding concrete type hint
    #   subscripting the passed generics and transitive pseudo-superclasses of
    #   that generic. Under this nomenclature, concrete type hints are seen as:
    #   * "Bubbling down" the DFS traversal tree from the passed generic rooting
    #     this tree into unpacked type parameters parametrizing the transitive
    #     pseudo-superclasses of that generic.
    #   * "Bubbling up" the transitive pseudo-superclass n-ary tree from the
    #     passed generic rooting this tree in the same manner. Since the natural
    #     depiction of a DFS is a DFS traversal tree, discussion below favours
    #     the term "bubbling down" to "bubbling up." The result is the same!
    #
    # And that's how @leycec got so exhausted in life. Not with an explosive
    # bang on GitHub Actions-hosted continuous integration (CI) workflows, but a
    # whimper as he quietly weeps into his tear-stained keyboard at night.
```


<sup>*throw those paws up in the air like you just don't bear*</sup>

leycec:
Thanks so much for inspiring my eventual hardcore resolution for #612, which required me refactoring literally the entire @beartype codebase. Understandably, even the usually brave LLM wasn't willing to go there. I almost wasn't, either. :joy: 


<sup>*the woman speaks the truth. what a hat, woman!*</sup>

binggao1230:
Glad it ended up useful as a stress test, even if the real fix needed the full refactor. #612's resolution looks like the right call — thanks for seeing it through.

leycec:
Thanks so much for your kind words, @gaoflow. Your generous PR volunteerism here spurred me to *finally* get off my bicycle and code something. The magic is all yours. :smile: 

Relatedly, [@beartype `0.23.0rc0` is now out with this fix](https://github.com/beartype/beartype/discussions/676) – and more! Like, *lots* more. Let us know if this new release solves all of your `is_subhint()` woes. If not, we will pound more things will hammers until they ~~shatter into a million irrecoverable pieces~~ finally start working again.

PR #677, opened 2026-08-28
Beardantic whispers: a `beartype.loadbearing` proof-of-concept

posita:
So there was [this thing that felt like something between a daydeam and an invitation](https://github.com/beartype/pytest-beartype/pull/33#issuecomment-5435983431) <sup>_which sounds like a Prince lyric_</sup> that opined whether plain type-hinted dataclasses could grow JSON Schema and marshalling superpowers via beartype.

At least that's what I came to realize after having Fabio explain it to me. That was right before he wrote pretty much all of this PR with a few gentle nudges <sup>_in the form of basic questions_</sup> from me.

What follows is mostly Fabio's, edited to compensate for his innate proclivities to periodically remind the reader of his semicolon-laden mechanical roots: an astounding command of language which is somehow devoid of anything resembling culture. (No offense, Fabio, buddy. You're killing it.)

_UPDATE: Busted tests should be addressed by #678._

---

This PR is a deliberately minimal proof-of-concept answering:
*probably, and DOOR is most of the reason why.* [It includes t]hree
commits[]:

1. **`to_json_schema()`** — a recursive walk over public
   `beartype.door` `TypeHint` trees emitting draft 2020-12 JSON
   Schemas.
2. **`to_json()` / `from_json()`** — structural serialization out,
   strict (coercion-free) deserialization in, with
   `die_if_unbearable()` as the final authority on well-typedness in
   *both* directions.

   _[He needed a nudge, but I think he's quite pleased with the
   result. --Ed.]_
3. **Differential tests against Pydantic** — `TypeAdapter` as oracle,
   compared modulo dialect. (Adds `pydantic` to the `test-tox` extra;
   it was already present transitively via `fastmcp`/`pandera`, but an
   oracle should be present on purpose. Tests skip gracefully where
   it's absent.)

   _[I have no idea what the first sentence fragment means, but I'm
   pretty sure the thrust of the third commit is to ask, "What would
   Pydantic do?" and then make sure we do something similar. --Ed.]_

### Supported subset (everything else raises `NotImplementedError`, loudly)

[Supported field types:] Scalars, `Any`, `Optional`/`Union`, `Literal`
(str/int/bool/None), scalar- valued `Enum`s,
`list`/`set`/`frozenset`/`dict[str, T]`, both tuple flavors, nested
non-recursive dataclasses. This covers the bulk of real LLM
tool-calling shapes while dodging every hard question on purpose.

### ~Headline result~ _[ :point_left:  We're not doing that. --Ed.]_

Over that entire subset, our schemas are **semantically identical to
Pydantic's**[. E]very difference is dialect (`$defs`/`$ref` vs
inlining, `title`/`default`/`description` decoration, redundant `type`
beside `const`/`enum`, unspecified orderings), all enumerated and
normalized in the differential test. ~350 lines of DOOR walk reproduce
the oracle on the subset that matters.

_[Why am I suddenly reminded of Hal Incandenza's university admissions
interview in_ Infinite Jest _? --Ed.]_

### DOOR substrate findings (the part [Fabio thinks] you'll actually care about)

* `TupleFixedTypeHint`/`TupleVariableTypeHint` are runtime-importable
  from `beartype.door` but missing from its explicit exports — mypy
  objects. `# type: ignore` + FIXME at the import sites; the upstream
  fix is a one-liner.
* `TypeHint.__iter__` is annotated `Iterable` rather than `Iterator`,
  so mypy rejects iterating a hint. Worked around by recursing via
  `.args` + `TypeHint(arg)` (arguably cleaner anyway).
* Otherwise DOOR held: subclass dispatch + `.args`/`.hint` covered the
  whole subset with zero private-API reaching. The PoC resolves hints
  via `typing.get_type_hints()`[. T]he real implementation should
  presumably use beartype's own resolution machinery instead.

### Policy decisions (hard-coded and documented, awaiting your taste)

* **`to_json()` fails loudly on hint-violating fields** (attribute
  mutation is unchecked, so latent violations are real). Rationale:
  serialization and deserialization are typically far apart[. A]
  silently serialized violation surfaces at some distant consumer
  where the source is untraceable. Alternatives (serialize faithfully;
  warn-and-emit à la Pydantic) are documented in-code as an open
  question.
* `from_json()` is strict on values, lax on shape — no coercion
  (including no bool-as-int), except int→float per JSON's numeric
  indifference[. U]nknown keys ignored. Notably this position has *no*
  Pydantic equivalent[. T]heir strict mode rejects dict input for
  dataclasses outright.
* Defaults ⇒ optional; `additionalProperties` left unset; nested
  dataclasses inlined (recursion raises rather than `$defs`/`$ref`,
  for now).
* Public signatures use the stub-only `_typeshed.DataclassInstance`
  protocol under `TYPE_CHECKING`. The beartype-native alternative —
  `beartype.vale` validators plus `@beartype`-decorating the API
  itself — is noted as a taste question for you.

### Deliberately not here (yet)

* The `loadbearing` attachment to `@beartype` itself — these are
  module functions; the decorator-attached namespace is sugar for a
  later commit, designed under your eye.
* `to_gbnf()` — the honest version chains our schemas through
  llama.cpp's `json_schema_to_grammar.py` rather than writing a
  grammar compiler; demoable next.

  _[Not sure if that means we have a dishonest version now, but I've
  found LLMs_ **love** _to talk about honesty, especially theirs.
  There's a line from MacBeth that might be apropos. --Ed.]_
* Stolen-from-Pydantic nicety worth adding: `description` derived from
  the dataclass docstring (the differential tests found this one).
* A survey of prior art (apischema, msgspec, cattrs)[. T]he sharper
  pitch isn't that schema-from-dataclass is novel (it isn't), but that
  beartype already owns the hard parts: hint resolution, DOOR, and the
  validator.

🤖 Generated with [omitted]  _[ :point_left:  Yeah, we're not doing that, either. --Ed.]_

### DIFF
--- a/beartype/loadbearing/__init__.py
+++ b/beartype/loadbearing/__init__.py
@@ -1,0 +1,17 @@
+#!/usr/bin/env python3
+# --------------------( LICENSE                            )--------------------
+# Copyright (c) 2014-2026 Beartype authors.
+# See "LICENSE" for further details.
+
+'''
+**Beartype load-bearing API** (i.e., public callables deriving portable
+JSON-centric artifacts -- including JSON Schemas -- from type-hinted
+:pep:`557`-compliant dataclasses via runtime introspection).
+
+This subpackage is currently a proof-of-concept exercising a deliberately
+minimal subset of type hints. See the :func:`.to_json_schema` docstring for the
+exact subset currently supported.
+'''
+
+# ....................{ IMPORTS                            }....................
+from beartype.loadbearing._schema import to_json_schema
--- a/beartype/loadbearing/__init__.py
+++ b/beartype/loadbearing/__init__.py
@@ -14,4 +14,8 @@ exact subset currently supported.
 '''
 
 # ....................{ IMPORTS                            }....................
+from beartype.loadbearing._marshal import (
+    from_json,
+    to_json,
+)
 from beartype.loadbearing._schema import to_json_schema
--- a/beartype/loadbearing/_marshal.py
+++ b/beartype/loadbearing/_marshal.py
@@ -1,0 +1,427 @@
+#!/usr/bin/env python3
+# --------------------( LICENSE                            )--------------------
+# Copyright (c) 2014-2026 Beartype authors.
+# See "LICENSE" for further details.
+
+'''
+**Beartype JSON marshallers** (i.e., low-level callables serializing instances
+of type-hinted :pep:`557`-compliant dataclasses to JSON-encodable dictionaries
+and deserializing such dictionaries back into validated instances).
+'''
+
+# ....................{ IMPORTS                            }....................
+#FIXME: See the sibling "_schema" submodule for commentary on the
+#"type: ignore" below. *sigh*
+from beartype.door import (  # type: ignore[attr-defined]
+    AnyTypeHint,
+    ClassTypeHint,
+    LiteralTypeHint,
+    SubscriptedTypeHint,
+    TupleFixedTypeHint,
+    TupleVariableTypeHint,
+    TypeHint,
+    UnionTypeHint,
+    die_if_unbearable,
+)
+from dataclasses import (
+    MISSING,
+    fields as dataclass_fields,
+    is_dataclass,
+)
+from enum import Enum
+from typing import (
+    TYPE_CHECKING,
+    get_type_hints,
+)
+
+# If statically type-checking, import the stub-only "DataclassInstance"
+# protocol. See the sibling "_schema" submodule for commentary.
+if TYPE_CHECKING:
+    from _typeshed import DataclassInstance
+
+# ....................{ SERIALIZERS                        }....................
+def to_json(datainst: 'DataclassInstance') -> dict:
+    '''
+    JSON-encodable dictionary serialized from the passed :pep:`557`-compliant
+    dataclass instance.
+
+    This proof-of-concept serializer is **structural** rather than
+    hint-driven: values are serialized according to their actual runtime types
+    rather than the type hints annotating their fields. Notably:
+
+    * :class:`enum.Enum` members serialize as their values (*not* names).
+    * Tuples, sets, and frozen sets serialize as JSON arrays (JSON defines
+      *no* tuple or set types). Set iteration order is unspecified; the
+      resulting array ordering is thus unspecified as well.
+    * Nested dataclasses serialize as nested JSON objects.
+
+    Before serializing, however, *all* fields of the passed instance are
+    validated against the type hints annotating those fields by
+    :func:`beartype.door.die_if_unbearable`. Fields of dataclass instances are
+    *not* reliably validated at assignment time (e.g., attribute mutation
+    after instantiation is unchecked even under :func:`beartype.beartype`), so
+    a hint-violating field is entirely possible here -- and would otherwise
+    silently serialize into JSON violating the corresponding
+    :func:`.to_json_schema` schema, deferring discovery to some arbitrarily
+    distant (and arbitrarily confusing) downstream consumer.
+
+    Parameters
+    ----------
+    datainst : DataclassInstance
+        :pep:`557`-compliant dataclass instance to be serialized.
+
+    Returns
+    -------
+    dict
+        JSON-encodable dictionary serialized from this instance.
+
+    Raises
+    ------
+    TypeError
+        If the passed object is *not* a dataclass instance.
+    beartype.roar.BeartypeDoorHintViolation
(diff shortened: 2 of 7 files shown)


JWCS:
I'm not exactly following what's all going on in this MR, or what the full goals of it are, but it slightly seems like adding serializers to beartype?
FWIW I've been using dataclasses and mashumaro (to avoid pydantic infecting all my types) (and because pydantic used to be slow) to do all my strongly typed beartyped dataclass serialization work, which has been pretty nifty (minus a few MREs after the next version drops). 
<details>
<summary>Given the following helpers:</summary>

```python
# std_ext_dataclasses.py
from pathlib import Path
from mashumaro.mixins.orjson import DataClassORJSONMixin

class DcJsonMixin(DataClassORJSONMixin):
    """Persistable dataclass: `to_json`/`from_json` from mashumaro + `save`/`fromfile`."""

    def save(self, path: str | Path) -> None:
        path = Path(path)
        path.parent.mkdir(parents=True, exist_ok=True)
        _ = path.write_text(self.to_json())

    @classmethod
    def fromfile(cls, path: str | Path):
        return cls.from_json(Path(path).read_text())

# std_ext_json.py
from collections.abc import Mapping, Sequence
from typing import TypeAlias

# TODO(followup): mashumaro-style typed serialization classes to replace ad-hoc dict[str, object] payloads.
JSON_T: TypeAlias = dict[str, "JSON_T"] | list["JSON_T"] | str | int | float | bool | None
JSON_ro: TypeAlias = Mapping[str, "JSON_ro"] | Sequence["JSON_ro"] | str | int | float | bool | None

JSON_ObjT: TypeAlias = dict[str, JSON_T]
JSON_ObjT_ro: TypeAlias = Mapping[str, JSON_ro]

# fs_schema.py # Does a bunch of other stuff wrt strongly typed filesystem schemas # That I ought to publish
...

FileBody: TypeAlias = bytes | str | Path

@runtime_checkable
class HasSave(Protocol):
    def save(self, path: Path) -> object: ...

@runtime_checkable
class HasSaveJson(Protocol):
    def save_json(self, path: Path) -> object: ...

Puttable: TypeAlias = FileBody | HasSave | HasSaveJson

def put(path: Path, data: Puttable) -> None:
    """Write one file; mkdir parents. `FileBody` or Protocol `.save` / `.save_json`."""
    path = Path(path)
    path.parent.mkdir(parents=True, exist_ok=True)
    match data:
        case HasSave():
            data.save(path)
        case HasSaveJson():
            data.save_json(path)
        case bytes():
            _ = path.write_bytes(data)
        case str():
            _ = path.write_text(data)
        case Path():
            shutil.copyfile(data, path)
        case _:
            raise TypeError(f"put: unsupported {type(data).__name__}")
```
</details>
The rest of our prod code mostly just types itself.
Ex, on the main dataclass we tend to serialize, mixin there (preferably with slots and frozen if one could). 
```python
@dataclass(frozen=True, slots=True)
class ApiType(DcJsonMixin):
    other_dataclass: MyType
    other_data: data
```

All this to say, while beartype is lacking some features (some things are not currently typechecked) for full typing quality, most stuff needed for strongly typed serialization is currently bolt-on...

posita:
> I'm not exactly following what's all going on in this MR, or what the full goals of it are, but it slightly seems like adding serializers to beartype? FWIW I've been using dataclasses and mashumaro (to avoid pydantic infecting all my types) (and because pydantic used to be slow) to do all my strongly typed beartyped dataclass serialization work, which has been pretty nifty (minus a few MREs after the next version drops).
> 
> ...
> 
> All this to say, while beartype is lacking some features (some things are not currently typechecked) for full typing quality, most stuff needed for strongly typed serialization is currently bolt-on...

I don't want to speak for @leycec, but [Mashumaro](https://mashumaro.io/) looks neat! Unless I'm mistaken, though, it looks like it requires a mix-in to enable serialization. Perhaps not. I haven't looked closely. My understanding of this experiment is: Can we do it **_without_** requiring authors to inherit from ... well ... anything?

We don't need no stinkin' mix-ins!

Could we instead leverage `beartype`'s existing type introspection such that the serialization/deserialization last mile is ... erm ... shorter ... than a mile? <sup>_That metaphor doesn't really work, but hopefully you get what I mean._</sup> If so, is something like `beartype_this_package` sufficient to get all that goodness? Turns out, at least for this limited-to-basic-builtins proof-of-concept, that "mile" is a couple dozen lines amidst a sea of (hopefully useful) comments. Maybe.

Is it fast? I don't know. Is it good? I ain't certain. [All that I know is we are on our way....
](https://youtu.be/m7YhYuS0RnU)

PR #680, opened 2026-09-03
Guard import-hook state against recursive loading

### LINKED ISSUE #640: [Feature Request] Future-proof `beartype.claw` against import cycles with "exciting" and "compelling" new `beartype._metaverse` submodule 💪 🐻
## Summary

beartype 0.22.9 hits the same `_clawimpload.py:317 → _clawstate` circular import as #638, but under a different trigger: **Temporal workflow worker validation** of a workflow module that transitively imports a beartype-using package (`py-key-value-aio` via `fastmcp`).

The fix on `main` for #638 is not in any released tag yet (0.22.9 is still latest on PyPI as of 2026-05-07), so production deploys hitting this path keep breaking.

## Error

```
File ".../beartype/claw/_importlib/_clawimpload.py", line 317, in get_code
    from beartype.claw._clawstate import claw_state
ImportError: cannot import name 'claw_state' from partially initialized module
'beartype.claw._clawstate' (most likely due to a circular import)
(/app/.venv/lib/python3.13/site-packages/beartype/claw/_clawstate.py)

The above exception was the direct cause of the following exception:

File ".../temporalio/worker/_workflow.py", line 168, in __init__
    raise RuntimeError(f"Failed validating workflow {defn.name}") from err
RuntimeError: Failed validating workflow W
```

## Environment

- Python 3.13.7 (Linux container, `python:3.13.7-slim`)
- beartype 0.22.9
- fastmcp 3.2.4
- py-key-value-aio 0.4.4 (transitive via fastmcp)
- temporalio 1.16.0

## Minimal repro

`pyproject.toml`:
```toml
[tool.poetry.dependencies]
python = "3.13.7"
fastmcp = "3.2.4"
temporalio = {version = "1.16.0", extras = ["opentelemetry"]}
```

`wf.py`:
```python
from temporalio import workflow
with workflow.unsafe.imports_passed_through():
    import fastmcp  # pulls py-key-value-aio → beartype.claw

@workflow.defn
class W:
    @workflow.run
    async def run(self) -> str:
        return "ok"
```

`run.py`:
```python
import asyncio
from temporalio.testing import WorkflowEnvironment
from temporalio.worker import Worker
from wf import W

async def main():
    async with await WorkflowEnvironment.start_local() as env:
        async with Worker(env.client, task_queue="tq", workflows=[W]):
            print(await env.client.execute_workflow(W.run, id="wid", task_queue="tq"))

asyncio.run(main())
```

Run inside `python:3.13.7-slim` container. Worker boot → `RuntimeError: Failed validating workflow W` caused by `claw_state` circular import.

## Observations

- Cold `import key_value.aio` succeeds.
- Cold `from beartype.claw._clawstate import claw_state` succeeds.
- Temporal sandbox importer (`temporalio.worker.workflow_sandbox._importer.Importer`) used outside a Worker also succeeds.
- Bug only triggers when the actual `Worker(workflows=[...])` constructor runs `_workflow.py:168` validation, which reloads the workflow module under sandbox restrictions.
- Pre-importing `beartype.claw` at the parent process level **does not** prevent the failure — sandbox validation reloads modules in a fresh namespace.

## Asks

1. Cut a 0.22.10 (or 0.23.0) release containing the #638 fix commits (44705c6, 2877b46, 578c633), since the fix landed on `main` 2026-04-30 but no release ships it.
2. Confirm whether that fix also covers this Temporal sandbox importer recursion path. If not, a separate guard in `_clawimpload.py:317` (or its caller) is likely needed.

Happy to package the repro as a Dockerfile if helpful.

brianstrauch:
## Summary

- track when the source loader is lazily importing its claw state
- bypass beartype transformation for nested loader calls until that state finishes initializing
- retain the package-name blacklist for known problematic packages while clarifying how the two guards differ

## Why

FastMCP transitively installs the beartype import hook during pytest collection. When a Temporal Worker later validates an unrelated workflow in its isolated module namespace, that process-wide hook can re-enter the source loader while beartype state is only partially initialized. The nested call then imports that state again and raises a circular ImportError.

The existing package blacklist does not cover this path because the re-entry can begin with an arbitrary module name. This change activates a context-local guard before importing hook state, so every nested loader call temporarily delegates directly to the standard source loader.

Fixes #640.

## Validation

- Beartype unit suite, Python 3.14: 420 passed, 10 skipped
- Beartype unit suite, Python 3.10: 407 passed, 23 skipped
- mypy: no issues in 474 source files
- temporalio/samples-python#358 full suite, Python 3.14: 170 passed, 4 skipped
- temporalio/samples-python#358 full suite, Python 3.10: 155 passed, 10 skipped

### DIFF
--- a/beartype/claw/_importlib/_clawimpfileloader.py
+++ b/beartype/claw/_importlib/_clawimpfileloader.py
@@ -15,6 +15,7 @@ This private submodule is *not* intended for importation by downstream callers.
 
 # ....................{ IMPORTS                            }....................
 from ast import PyCF_ONLY_AST
+from contextvars import ContextVar
 from beartype.roar._roarexc import (
     _BeartypeClawImportlibIsPathHookActiveException)
 from beartype.claw._ast.clawastmain import BeartypeNodeTransformer
@@ -38,6 +39,13 @@ from importlib.machinery import SourceFileLoader
 from importlib.util import decode_source
 from types import CodeType
 from typing import Optional
+
+# True only while this module is lazily importing the state required to decide
+# whether the current module should be transformed. Third-party importers may
+# recursively invoke this loader while those imports are still in progress.
+_IS_GET_CODE_IMPORTING_STATE: ContextVar[bool] = ContextVar(
+    '_IS_GET_CODE_IMPORTING_STATE', default=False)
+
 
 # ....................{ CLASSES                            }....................
 #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
@@ -336,6 +344,13 @@ class BeartypeSourceFileLoader(SourceFileLoader):
             efficiently detect whether they were successfully activated or not.
         '''
 
+        # ..................{ RECURSE                        }..................
+        # If a lazy import below recursively invokes this loader, preserve the
+        # nested module as is. Attempting to import our state again would recurse
+        # into a partially initialized "beartype" package.
+        if _IS_GET_CODE_IMPORTING_STATE.get():
+            return super().get_code(fullname)
+
         # ..................{ SMOKE                          }..................
         # If that module name is that of the beartype import hook activation
         # smoke test, raise the exception expected by that test. *YO*!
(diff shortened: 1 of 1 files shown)


brianstrauch:
Confirmed that upstream commit f593cd6f fixes the Temporal samples without this patch. The exact 13-test reproducer, including the unrelated Worker tests and all Pydantic AI plugin samples, passes on both Python 3.10 and Python 3.14. Closing as superseded by the upstream fix.

PR #681, opened 2026-09-04
Handle the bare UnionType class in get_hint_pep_childs

### LINKED ISSUE #609: [Bug] `is_subhint(..., type[Union])` busted under Python ≤ 3.13 😞
In python 3.14,
```python
>>> is_subhint(type(int | str), Union)
True
>>> is_subhint(type[int | str], Union)
False
>>> is_subhint(type[int | str], type[Union])
True
>>> is_subhint(int | str, Union) # should be True?
False
```
I "feel" like that `is_subhint(type[int | str], type[Union])` and `is_subhint(int | str, Union)` need to be consistent. So both are true or both are false. However, I have no strong reasoning for this. If beartype interprets `Union` as "any union expression", should both be `True`? If the current behavior is intendend (or consistent with correspondig PEP), I think it is OK. I just want to know whether it is intended or a bug.

In python 3.13,
```python
>>> is_subhint(type(int | str), type[Union])
...
beartype.roar.BeartypeDecorHintPepException: PEP-noncompliant hint <class 'types.UnionType'> "__parameters__" dunder attribute value <attribute '__parameters__' of 'types.UnionType' objects> invalid (i.e., not tuple of PEP 484-, 612-, or PEP 646-compliant type parameters).
>>> is_subhint(type[int | str], type[UnionType])
...
beartype.roar.BeartypeDecorHintPepException: PEP-noncompliant hint <class 'types.UnionType'> "__parameters__" dunder attribute value <attribute '__parameters__' of 'types.UnionType' objects> invalid (i.e., not tuple of PEP 484-, 612-, or PEP 646-compliant type parameters).
>>> is_subhint(int | str, Union)
True
>>> is_subhint(type[int | str], Union)
True
>>> is_subhint(type[int | str], type[Union])
True
```
For python< 3.14, `is_subhint` does not work consistenly with `Union` or `UnionType`. Would you support `types.UnionType` or `Union` for python < 3.14?

In python 3.14,
```python
>>> is_subhint(type(int), Union)
False
>>> is_subhint(type[int], type[Union]) # should be False?
True
```
I'm not sure that regarding a single type as a kind of union is a valid interpretation. I'm implemening a factory function taking  a type hint as an argument. The factory needs to consider the hierarchy of types and has a specialized logic for unions. So, I need to distinguish a single type and a union of types. If this behavior is not intended, would you fix? Thanks!

afonsojanu:
Fixes #609.

`is_subhint(type(int | str), type[Union])` and similar calls crash with a `BeartypeDecorHintPepException` on Python < 3.14. The root cause is in `get_hint_pep_childs()`: when the bare `types.UnionType` class itself turns up as a hint (rather than an actual PEP 604 union instance like `int | str`), reading its `__args__` gets you the class's inherited unbound slot descriptor instead of a real tuple, and the function raises.

There's already a special case right above this one handling the exact same shape of problem for bare `typing.Union` on Python 3.14+, where `Union` became a C-based type with the same kind of non-tuple `__args__`. I just extended that same check to also cover `types.UnionType`, so the bare class is treated as having no children rather than blowing up.

Added two cases to the `is_subhint` fixture covering both the `type(int | str)` and `type[int | str]` shapes from the issue. Confirmed both crash on the unpatched code and pass with the fix (via `git stash`). Ran the full local suite: 422 passed, 41 skipped (all skips are pre-existing, due to optional dependencies not installed here, like numpy/torch/pandas).

### DIFF
--- a/beartype/_util/hint/pep/utilpepget.py
+++ b/beartype/_util/hint/pep/utilpepget.py
@@ -46,6 +46,7 @@ from beartype._util.hint.pep.proposal.pep585 import (
 )
 from beartype._util.hint.pep.proposal.pep646.pep646692unpack import (
     make_hint_pep646_typevartuple_unpacked_subbed)
+from types import UnionType
 from typing import (
     Any,
     Optional,
@@ -158,7 +159,16 @@ def get_hint_pep_childs(hint: object) -> tuple:
         # valid hints, this "__args__" implementation is *TECHNICALLY* also
         # valid albeit semantically meaningless. In this case, simply return the
         # empty tuple.
-        if hint is Union:
+        #
+        # Likewise, if this hint is the unsubscripted "types.UnionType" class
+        # itself (as opposed to an actual PEP 604-compliant union instance
+        # like "int | str"), this hint's "__args__" dunder attribute is
+        # merely the unbound slot descriptor inherited from the class, not an
+        # actual tuple of child hints. This arises whenever a caller passes
+        # "type(some_union)" or "UnionType" itself as a hint (e.g., inside a
+        # "type[...]" hint). As with the "Union" case above, return the empty
+        # tuple rather than erroring out.
+        if hint is Union or hint is UnionType:
             return ()
         # Else, this hint is *NOT* the unsubscripted "typing.Union" hint. In
         # this case, raise an exception.
--- a/beartype_test/a00_unit/a40_api/door/_fixture/_doorfix_is_subhint.py
+++ b/beartype_test/a00_unit/a40_api/door/_fixture/_doorfix_is_subhint.py
@@ -107,6 +107,7 @@ def door_cases_is_subhint() -> 'Iterable[Tuple[object, object, bool]]':
         TypedDict,
         Union,
     )
+    from types import UnionType
 
     # ..................{ NEWTYPES                           }..................
     NewStr = NewType('NewStr', str)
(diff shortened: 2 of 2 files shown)


leycec:
**Guh!** So sorry for the extreme delay, @afonsojanu. I... somehow totally missed this awesome PR. That's entirely my fault, obviously. Unfortunately, *because* I missed this, there are now merge conflicts. I sigh. :face_exhaling: 

If you'd like to resolve those merge conflicts, that would be wonderful! Otherwise, I might have a go at them myself. It's super-annoying that this is *still* a issue. If one of us fixes this in the next few days, this issue resolution has an excellent chance of landing in @beartype `0.23.0` – which absolutely *must* be released next week, because CPython 3.15.0 drops the week after.

*Anywaaaays.* Thanks so much for this. You're awesome, fellow QAthonista!

leycec:
**Manually merged at 8c93d65513693a2b.** Thanks so much for your super-heroic volunteerism here, @afonsojanu. I manually resolved all merge conflicts and... tests were still broken under Python ≤ 3.13. Looks like you forgot to run `tox`, huh? It happens. I myself feel the strong pull of video games and laziness. :joy: 

So, I resolved all the trivial test failures under Python ≤ 3.13 too. Looks like everything's good to go now. We did it, everybody! We did something. Issue #609 has been crushed. This issue resolution barely made the cut for our upcoming @beartype `0.23.0` release – but made it, it did. My emoji is sweating. :sweat_smile:

Issue #689, opened 2026-09-15
[DOOR Feature] `beartype.door.TypeHint` support for PEP 695-compliant `type` aliases = 🎃 


Glinte:
```py
from beartype.door import TypeHint

type X = int
TypeHint(X)
```

```txt
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\user\AppData\Local\uv\cache\archive-v0\51rG_u6huwoFxOJU\Lib\site-packages\beartype\door\_cls\doormeta.py", line 133, in __call__
    _hint_to_wrapper.cache_func_return_passed_arg(  # type: ignore[assignment]
  File "C:\Users\user\AppData\Local\uv\cache\archive-v0\51rG_u6huwoFxOJU\Lib\site-packages\beartype\_cache\cls\cacheclsmega.py", line 398, in cache_func_return_passed_arg
    value_new = value_factory(arg)
                ^^^^^^^^^^^^^^^^^^
  File "C:\Users\user\AppData\Local\uv\cache\archive-v0\51rG_u6huwoFxOJU\Lib\site-packages\beartype\door\_cls\doormeta.py", line 211, in _make_wrapper
    wrapper_subclass = get_typehint_subclass(hint)  # pyright: ignore
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\user\AppData\Local\uv\cache\archive-v0\51rG_u6huwoFxOJU\Lib\site-packages\beartype\door\_cls\_doormap.py", line 133, in get_typehint_subclass
    raise BeartypeDoorNonpepException(
beartype.roar.BeartypeDoorNonpepException: Type hint X currently unsupported by "beartype.door.TypeHint".
```

leycec:
**...ho, ho.** Yes! Indeed, it'd be fun to eventually add support for PEP 695-style `type` aliases to the `beartype.door.TypeHint` API. Maybe *not* a huge priority? If plum (or somebody else) has an active and growing hunger for this, just let us know and we'll reprioritize accordingly. My laziness is legendary and knows no bounds.

In theory: this *should* be a trivial add. In practice: nothing is trivial, everything is scary, and the seasonally appropriate Jack-o'-lantern best exemplifies this feature request.

![](https://media1.tenor.com/m/t5DYo7jiGXAAAAAC/love-live-school-idol-project.gif)
[...]

Issue #626, opened 2026-03-20
[Feature Request] Leycec's Last Theorem of QA: "Resolve forward references to undefined arbitrarily complex type hints, now!" :finnadie: 


ilyapoz:
Hi, leycec!

I'm happy and sad at the same time. 
Happy, because I'm once again conversing with you.

Sad for 3 reasons:
1. bug in beartype
2. I addressed you by name in an open source issue, being sad that I assume you're the only one reacting to this. I really wish you hordes of helpful supporters&maintainers for your awesome project, man!
3. I forgot the third one.

This issue is similar to  https://github.com/beartype/beartype/issues/577 (just my guess).

The reproducer

```python
% cat main.py
#!/usr/bin/env -S uv run
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "beartype==0.22.9",
# ]
# ///

import beartype

@beartype.beartype
def my_fn(x: "friend.Great"):
    print(f'hey, {x}')

my_fn(42)
```

```python
% cat friend.py
class X:
    pass

type Great = dict[str, X]
```

```
% ./main.py
...
beartype.roar.BeartypeCallHintForwardRefException: Forward reference Great neither class nor tuple of one or more classes.
```
👆
 Man, I love this PEP magic + uv combo. I now use this everywhere!

If I remove the `type` prefix (I thought it was cool):

```
% cat friend.py
class X:
    pass

# TODO: add type
Great = dict[str, X]
```

```
% ./main.py
...
beartype.roar.BeartypeCallHintForwardRefException: Forward reference dict[str, friend.X] neither class nor tuple of one or more classes.
```
The error is silghtly and nontrivially different 🤔 I'm not qualified to make any implications though. Ok, maybe just that my syntax must be valid!

I'm not a great standards lawyer like the reporter of the last issue, so I'm not sure my case is covered by THE S. But still, putting this out there for your judgement, your honor 🙇  😁

Glinte:
Have you tried the lastest commit? leycec reworked the whole forward ref resolution mechanism a few weeks ago

ilyapoz:
Thanks!
I tried
```diff
 # dependencies = [
-#     "beartype==0.22.9",
+#     "beartype @ git+https://github.com/beartype/beartype",
 # ]
```

And it gives the same error

leycec:
[...]

PR #691, opened 2026-09-16
Halt on circular PEP 695 type alias chains

nstarman:
## Problem

A circular chain of PEP 695 type aliases hangs the interpreter — forever, with no exception and no `RecursionError`:

```py
type A = B
type B = A

is_bearable(1, A)   # never returns
```

`get_hint_pep695_unsubbed_alias()` iteratively unwraps nested aliases in an unguarded `while True:` loop. Each iteration finds another alias, so the loop never breaks.

This is reachable from `@beartype` and `is_bearable()` — not merely from `beartype.door`.

## Distinction

A circular chain is *not* the same as a **recursive** alias:

```py
type Tree = int | list[Tree]      # recursive: valid, conveys real semantics
type A = B; type B = A            # circular: conveys no semantics at all
```

Recursive aliases are untouched by this change and keep working exactly as before. A circular chain, by contrast, never reduces to a hint conveying anything — unwrapping it only ever yields another alias in the same chain.

## Fix

Guard the loop with the set of aliases already unwrapped by this call and raise a human-readable `BeartypeDecorHintPep695Exception` on revisiting one:

```
BeartypeDecorHintPep695Exception: ... PEP 695 type alias A circularly aliases
itself, thus conveying no semantics. Consider breaking this circularity: ...
```

## Tests

`test_get_hint_pep695_unsubbed_alias_circular` covers self-circular (`type X = X`) and two-alias chains, gated on Python >= 3.12 following the existing `IS_PYTHON_AT_LEAST_3_12` convention. The aliases live in `data_pep695hint.py` with the other PEP 695 test data.

`424 passed, 9 skipped` on `beartype_test/a00_unit` (3.13.14).

## Context

Found while implementing #689 (`beartype.door.TypeHint` support for PEP 695 aliases). That work inherits this hang, so it is split out here as an independent fix — it stands on its own and needs no DOOR changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Disclaimer: this one was ALL Claude. Once we identified the bug I just asked it to fix it and open a PR. I only eyeballed the code looking for anything egregious. I will not be offended by asks for major rewrites, esp verbosity, which appears to be LLM-standardly-too-much.


leycec:
**...woah.** Superb catch, you two. Circular `type` alias chains are pathological, which probably explains why no one ever hit this before. A user would almost have to go out of their way to intentionally synthesize one of these monstrosities.

A horrifying bug is still a bug, though. Since this bug is obviously horrifying, this is now a hard blocker on @beartype `0.23.0`. That's fine. I can deal with this. I can deal... `</starts_huffing_into_paper_bag>`


<sup>*that fatal moment when you realize you can no longer deal with something*</sup>

## In All Seriousness...

Thanks so much for both detecting this *and* proposing the awesome solution. :partying_face: 

The use of object IDs (e.g., `id(hint)`) makes me squirm a bit. Everyone *thinks* that object IDs are unique – apparently including Claude. They're not. CPython (and presumably other Python interpreters) recycle object IDs. It's almost always a bad idea to use them, inviting subtle issues with both caching and thread concurrency. That said...

This is *probably* one of the few cases where it's actually safe! The object ID lookup is confined to a single callable introspecting a single immutable object (namely, a type hint). I'm pretty sure this means that:

* Both you and Claude did an awesome spot of open-source volunteerism here.
* I'll merge this as is tomorrow night.

To quote the venerable Meatloaf, "Let me sleep on it." :joy:

leycec:
Thanks for your continued awesomeness, fellow brother in QA. This was *quite* a critical issue. Any type hint that induces infinite recursion in @beartype (or other runtime type-checkers) could be considered a potential DoS and thus security risk from a certain perspective. My avatar is truly sweaty tonight. :hot_face:

Glinte:
I didn't test it, does this behavior not exist for normal forward references like `A = "B"; B = "A"`? This looks like a `TypeAliasType`-specific fix

leycec:
**Yuppers!** Excellent catch. Indeed, the same exact issue arises with both:

* PEP 484-compliant stringified circular forward references like `A = "B"; B = "A"`.
* PEP 649- and 749-compliant unquoted circular forward references like `A = B; B = A`.

I dimly recall implementing preliminary guards against both a few months back. Now, where did that code run off too... Oh! Here it is. Deep in the festering innards of `fwdrefcache`:

```python
            # If this referent is this forward reference proxy, this proxy
            # circularly proxies itself. Since allowing this edge case would
            # openly invite infinite recursion, we detect this edge case and
            # instead raise a human-readable exception.
            if referent_hint is ref_proxy:
                raise BeartypeCallHintForwardRefException(
                    f'{_make_ref_proxy_exception_prefix(ref_proxy)}'
                    f'that target referent circularly '
                    f'(i.e., infinitely recursively) references itself.'
                )
            # Else, this referent is *NOT* this forward reference proxy.
```

No idea whether that detection covers only trivial circularity (e.g., `A = 'A'`) or both trivial and non-trivial circularity (e.g., `A = 'B'; B = 'A'`). Can't remember, honestly. Too bad I didn't document it properly. Hopefully, at least something is tested... *Hopefully.*

Edge cases like this are why @ilyopoz's long-standing forward reference feature request #626 is still no closer to being closed out. I kinda just gave up at a certain point. :sob:

PR #692, opened 2026-09-16
Support PEP 695 type aliases in beartype.door.TypeHint

### LINKED ISSUE #689: [DOOR Feature] `beartype.door.TypeHint` support for PEP 695-compliant `type` aliases = 🎃

nstarman:
Resolves #689.

```py
from beartype.door import TypeHint
type X = int
TypeHint(X)   # was: BeartypeDoorNonpepException
```

## Cause

`beartype/door/_cls/_doormap.py` maps signs to `TypeHint` subclasses and has no entry for `HintSignPep695TypeAliasUnsubscripted` / `…Subscripted`. `get_typehint_subclass()` therefore falls through its three fallbacks — `isinstance(hint, type)` is false (a `TypeAliasType` is not a class), `is_hint_pep_typing()` is false (it is a C-level object, not a `typing` member) — and raises.

Sign detection was always correct, so **`get_typehint_subclass()` is untouched** by this PR: the lookup now hits and simply never reaches the raising fallback. The `_doormap` diff is 4 purely additive lines.

Scope is narrower than the issue title suggests. PEP 695 *type parameters* already worked on `main` — `def f[T]`, `def f[T: int | float]` (bounds respected), `class C[T]`, `def f[*Ts]`, `def f[**P]`. Only `type X = …` aliases were broken, and only through DOOR; `is_bearable()` handled them all along.

## Design

`Pep695TypeAliasTypeHint` keeps the alias as `_hint` and defers every semantic decision to a lazily-built wrapper of the aliased hint. Three details carry it:

- **Laziness is load-bearing, not an optimization.** `@property_cached` rather than `__init__` is what makes `TypeHint(Tree)` terminate for `type Tree = int | list[Tree]`.
- **Subscripted aliases reuse CPython's own substitution** — `(list[T] | set[T])[int]`. No type-parameter lookup table is needed, which is what makes this tractable inside DOOR.
- **`_branches` delegates too.** `TypeHint._is_subhint` iterates `other._branches`, so without this an alias would be transparent as the left operand and opaque as the right, giving asymmetric equality.

### `TypeHint.unaliased`

`TypeHint(X) is TypeHint(int)` and `TypeHint(X).hint is X` cannot both hold. Rather than pick, this exposes the delegate — which already *was* the cached singleton:

```py
TypeHint(X).unaliased is TypeHint(int)    # True
TypeHint(X).hint is X                     # True
TypeHint(int).unaliased is TypeHint(int)  # True -- base property returns self
```

Callers never branch. This also recovers `isinstance(th.unaliased, ClassTypeHint)` for code introspecting by wrapper type.

**This property is severable** — happy to drop it or rename it if you would rather keep the API surface flat, or land it separately.

## This is "Plan B", and there is a "Plan B′"

I built two designs and benchmarked both. This PR is the one that adds a subclass. The alternative resolves aliases **transparently** in `_TypeHintMetaclass`, with no new class, so that `TypeHint(X)` *is* `TypeHint(int)` — one singleton, `isinstance(th, ClassTypeHint)` working directly, and no `.unaliased` needed.

I picked this one because:

- it touches no hot code (`doorabc.__hash__` / `is_subhint` are untouched; B′ needs recursion guards in all three of `__hash__`, `is_subhint`, and `__eq__`);
- it preserves the alias, so `.hint` and `repr` can still answer "what did the user actually write?" — B′ discards that irrecoverably;
- it matches the `door/__init__.py` FIXME asking for one subclass per kind of hint, for exactly that introspectability.

**The honest case for B′:** existing third-party code introspecting DOOR by wrapper class sees aliases transparently under B′ and meets an unfamiliar class under this PR. Nothing regresses either way (today such code gets an exception), but B′ makes alias support free for existing consumers.

**If you prefer B′, say the word and I will submit it instead** — it is already spiked and tested, so it is a quick turnaround.

## Performance

Interleaved A/B, medians of 11 reps, against a ~1–2% `main`-vs-`main` noise floor (3.13.14, ns/op):

| | main | this PR |
|---|---|---|
| `TypeHint(int)` | 582.2 | 580.2 |
| `TypeHint(int\|str)` | 890.1 | 895.3 |
| `hash(th)` | 65.3 | 65.8 |
| `th <= th` | 402.1 | 401.2 |
| `is_subhint(a,b)` | 1891 | 1921 |
| `is_bearable(1,int)` | 331.3 | 333.1 |

No regression — expected, since nothing on the hot path changed. New alias paths cost the same as their non-alias equivalents once warm (`TypeHint(alias)` 576 vs `TypeHint(int)` 580).

## Equality and hashing

Recursive aliases make the wrapper graph cyclic, so hashing and comparison need a guard, and the guard's *fallback value* decides whether `__eq__`/`__hash__` stay consistent. Returning `hash(self._hint)` on re-entry is wrong: `type Tree = int | list[Tree]` and an identical `type Tree2 = …` compare equal but hash differently, so `{TypeHint(Tree): 1}[TypeHint(Tree2)]` silently misses. This PR returns a shared module constant instead, which holds the constraint — including when the two aliases were first reached in different traversal orders. `test_door_pep695_alias_recursive` asserts it.

## Tests

- PEP 695 cases appended to both `_door_cases_equals.py` and `_door_cases_subhint.py`, gated on `IS_PYTHON_AT_LEAST_3_12` with the aliases in `data_pep695hint.py`, following existing convention.
- New `test_door_typehint_pep695.py` covering the wrapper, `.unaliased`, and recursive-alias hashing/comparison.

`426 passed, 9 skipped` on `beartype_test/a00_unit` (3.13.14).

## Known limitation, not introduced here

An alias of a *string* forward reference (`type Fwd = "Foo"`) still fails, because DOOR has no forward-reference support at all — `TypeHint("Foo")` fails on `main` too. Aliases merely give another route there. Filed as #694; this PR neither improves nor regresses it.

## Related

- #691 fixes a circular alias chain (`type A = B` + `type B = A`) hanging the interpreter. Independent of this PR and reachable from `@beartype`/`is_bearable()`, but this PR inherits that hang, so #691 is worth landing first.
- #693: `typing_extensions.TypeAliasType` breaks the *decorator* path — `HintPep695TypeAlias` in `_cave/_cavefast.py` binds only `typing.TypeAliasType`, so `reduce_hint_pep695_unsubbed()` asserts. Unrelated to DOOR, but it means PEP 695 support is uneven depending on where the alias came from.
- #694: `beartype.door.TypeHint` cannot wrap stringified forward references. Pre-existing and DOOR-wide; noted above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


leycec:
> This PR is the one that adds a subclass.

**Correct!** This is the way. You chose... wisely. Your fine discernment is a lesson to us all. :+1: 

> This PR returns a shared module constant instead, which holds the constraint...

**...uh oh.** That seems kinda weird and scary. I'm choosing to believe thread-unsafe global variables aren't involved here, Claude. Don't let us down now! :disappointed_relieved: 

Aaaaaand, I guess I'd better speed-read this changeset. What I put off until tomorrow can only get worse.


<sup>*truly the look of a determined bearded man with hair*</sup>

beartype/door/_cls/pep/pep695/doorpep695.py
```diff
@@ -0,0 +1,212 @@
...
+        :meth:`__init__` method. Why? Because :pep:`695` type aliases are
+        lazily evaluated and thus commonly recursive (e.g., ``type Tree = int |
+        list[Tree]``). Deferring this decision until required guarantees that
+        merely instantiating this wrapper is guaranteed to halt.
+        '''
+
+        # Unsubscripted type alias underlying this possibly subscripted type
+        # alias, defined as either this alias itself if unsubscripted *OR* the
+        # origin of this alias otherwise.
+        hint_unsubbed = getattr(self._hint, '__origin__', self._hint)
```
leycec: **...yikes.** Kinda awkward. Definitely unsafe. It's best to *never* directly touch low-level dunder attributes like `__origin__`, because they blow up at the slightest provocation. Instead, let's define a new high-level `get_hint_pep695_alias()` getter resembling:

```python
#FIXME: Unit test us up, someone. Please. We're begging here. *sigh*
@callable_cached
def get_hint_pep695_alias(
    # Mandatory parameters.
    hint: HintPep695TypeAliasTypes,

    # Optional parameters.
    exception_prefix: str = '',
) -> Hint:
    # Avoid circular import dependencies.
    from beartype._util.hint.pep.utilpepget import get_hint_pep_origin

    if is_hint_pep695_subbed(hint):
        hint = get_hint_pep_origin(hint)

    return get_hint_pep695_unsubbed_alias(
        hint=hint, exception_prefix=exception_prefix)
```

Muscly bear intensifies. 💪 🐻

beartype/door/_cls/pep/pep695/doorpep695.py
```diff
@@ -0,0 +1,212 @@
...
+
+        # Hint aliased by this alias. Note that this getter is memoized and thus
+        # intentionally called with positional arguments.
+        hint_aliased = get_hint_pep695_unsubbed_alias(hint_unsubbed, '')  # type: ignore[arg-type]
```
leycec: Aaaand... let's replace this assignment with a call to that higher-level `get_hint_pep695_alias()`:

```python
        hint_aliased = get_hint_pep695_alias(hint)
```

beartype/door/_cls/pep/pep695/doorpep695.py
```diff
@@ -0,0 +1,212 @@
...
+
+        # Child hints subscripting this alias if any *OR* the empty tuple.
+        hint_childs = get_hint_pep_childs(self._hint)
```
leycec: **Unneeded.** This tuple is already trivially available as `self._args`: e.g.,

```python
    hint_childs = self._args  # <-- silly but conveys the idea *shrug*
```

beartype/door/_cls/pep/pep695/doorpep695.py
```diff
@@ -0,0 +1,212 @@
...
+        return TypeHint(hint_aliased)
+
+
+    @property
+    def unaliased(self) -> TypeHint:
+
+        # Wrapper wrapping the hint aliased by this alias, which the caching
+        # "_TypeHintMetaclass" metaclass guarantees to be the same singleton
+        # wrapper returned by passing that hint to the "TypeHint" superclass.
+        return self._hint_aliased_wrapper
```
leycec: **...kinda weird.** It's fine if we want to publicly expose this read-only property to users. Makes sense. There's no reason to define two separate properties, though. Right? One suffices. Thus, let's just:

* Rename the `_hint_aliased_wrapper` property defined above to `hint_aliased`.
* Remove this redundant `unaliased` property. `hint_aliased` is a more descriptive name, anyway. `</shrugs_bro>`

beartype/door/_cls/pep/pep695/doorpep695.py
```diff
@@ -0,0 +1,212 @@
...
+    @property
+    def _is_args_ignorable(self) -> bool:
+        return self._hint_aliased_wrapper._is_args_ignorable
+
+    # ..................{ PRIVATE ~ factories                }..................
+    def _make_args(self) -> tuple:
+
+        # Preserve the child hints subscripting this alias (if any), enabling
+        # the len(), iter(), and [] APIs to behave as expected.
+        return get_hint_pep_childs(self._hint)
```
leycec: **...lol.** This is exactly the superclass implementation and thus totally redundant. Let's remove this entirely. Claude. C'mon, buddy. Try to keep up. *y u do this 2 us* 😮‍💨

beartype/door/_cls/pep/pep695/doorpep695.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # ..................{ PRIVATE ~ getters                  }..................
+    def _get_hash(self) -> int:
+
+        # Set of the IDs of all type aliases currently being hashed by the
+        # current thread. PEP 695 type aliases are commonly recursive (e.g.,
+        # "type Tree = int | list[Tree]"). Since hashing a recursive alias as
+        # the hint aliased by that alias would provoke infinite recursion,
+        # detect and halt that recursion here.
+        hint_ids = _hint_ids_hashing.ids
```
leycec: **NOPE.** There exists a one-to-one mapping between object identifiers and memory addresses in CPython. Since memory addresses are frequently reclaimed in long-running processes, object identifiers are frequently recycled in long-running CPython processes. Attempting to do *anything* with object identifiers across callable and thread boundaries is a fundamental non-starter.

It's likely that Claude has over-complicated this to an extreme degree. @beartype has *never* needed shadow madness like this before – and certainly isn't going to start now:

```python
_hint_ids_hashing = _HintIdsThreadLocal()  # <-- NO NO NO NO NO
_hint_ids_comparing = _HintIdsThreadLocal()  # <-- NOOOOOOOOOOO
```

🤮 🤮 🤮 

Because this is vomitous, it's late, and there's already a litany of other pressing issues, I stop here. This PR is well-intended but needs a fundamental rethink with respect to recursion guard protection.

Unfortunately, the only solution I can think of is... non-trivial. That solution resembles what the main @beartype codebase does to protect against infinite recursion. In this case:

* Let's define a new private `def get_hash(type_hint: TypeHint, type_hints_seen: frozenset[TypeHint]) -> int:` getter somewhere inside the private `beartype.door._cls` subpackage. Doesn't particularly matter where. I shrug. 🤷
* The signature probably resembles:

  ```python
  @callable_cached
  def get_hash(
      type_hint: TypeHint, type_hints_seen: frozenset[TypeHint]) -> int:
      ...
  ```

* Here's the non-trivial part. Actually, it's mostly just tedious. **Refactor all `TypeHint._get_hash()` subclass implementations** to call our new high-level `get_hash()` getter (which guards against recursion) rather than the low-level `hash()` builtin (which fails to guard against recursion, obviously).

**...yeah.** Let's all vomit as one, everybody! 🤮

leycec (commented):
**...heh.** Substantial issues abound, sadly. Looks like this won't make @beartype `0.23.0` after all. Oh, well. We tried! And in the end, isn't trying but failing all that matters? 🥹

beartype/door/_cls/pep/pep695/doorpep695.py
nstarman: Done in 689fff8f: `get_hint_pep695_alias()` now lives in `beartype/_util/hint/pep/proposal/pep695.py`, `@callable_cached` and shaped like your sketch (reduce a subscripted alias to its origin, then `get_hint_pep695_unsubbed_alias()`). No more `__origin__` in DOOR. Its `#FIXME: Unit test us up` has been answered with `test_get_hint_pep695_alias`, which covers unsubscripted, subscripted, nested, non-alias and circular inputs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

beartype/door/_cls/pep/pep695/doorpep695.py
nstarman: Done: `hint_aliased` calls `get_hint_pep695_alias(self._hint, '')`, positionally so the cache hits.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

beartype/door/_cls/pep/pep695/doorpep695.py
nstarman: Done. The `_make_args()` override below is removed too.

beartype/door/_cls/pep/pep695/doorpep695.py
nstarman: Done! It's one public property now.

beartype/door/_cls/pep/pep695/doorpep695.py
nstarman: Removed. Even Claude was abashed, saying 

> *y u do this 2 us* was fair. 🙇

beartype/door/_cls/pep/pep695/doorpep695.py
nstarman: Agreed, the thread-local `id()` sets are gone. There's no global, thread-local or `id()`-keyed state anywhere in the PR now.

Rather than add recursion guards across every `_get_hash()`, this round rejects **recursive** aliases outright and defers real support to #701. `TypeHint(Tree)` raises `BeartypeDoorNonpepException` at construction, the same exception every alias raised before this PR, so callers that catch it keep working. Detection is `is_hint_pep695_recursive()` in `_util`. It uses local sets only and runs in linear time, and it catches mutual recursion and recursion through another alias's subscription (`type Tree = int | Wrap[Tree]`). Non-recursive aliases need no guards at all.

On the `get_hash(type_hint, type_hints_seen: frozenset[TypeHint])` sketch, two snags for when #701 happens, both written up there:
- Putting wrappers into a `frozenset` hashes them, which is the very recursion being guarded, and `@callable_cached` would key on the same wrappers. The seen-set probably wants the raw alias objects instead, since every cycle has to pass through an alias.
- The hash has to be the same whichever node traversal starts from, because `TypeHint(Tree) == TypeHint(int | list[Tree])` requires equal hashes. Truncating at the first repeated alias cuts those two at different depths.

While here I also found that the alias wrapper leaked into subclass testers as the *other* operand, which made `TypeHint(list[int]) == TypeHint(Alias)` asymmetric. `__eq__`/`__ne__`/`is_subhint` now unwrap aliases before deferring, and unions build `_branches` from their children's `_branches` (dec1d371). The PR body has the details. Separately, `repr()`-keyed memoization collides on same-named aliases (and on today's same-named `TypeVar`s), which I've filed as #700.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

nstarman:
Heads-up: I rebased this branch onto `main` at 0d760a1b, which replaced your "Update branch" merge commit (nothing from it was lost).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

